PR #197, opened 2020-06-30
allow to query ARP by looking at the known mac address

anubisg1:
allow to query ARP by looking at the known mac address


anubisg1:
allow to query ARP by looking at the known mac address

Signed-off-by: Andrea Florio andrea@opensuse.org

PR #229, opened 2020-07-28
add "fs" support to suzieq-cli

anubisg1:
add "fs" support to suzieq-cli

Signed-off-by: Andrea Florio andrea@opensuse.org


ddutt:
@anubisg1  Thank you for this PR. I have a few comments. Could you please fix these and resubmit. I realize that one of these breaks the tests and you can't fix that. That's OK.

ddutt:
@anubisg1 The changes are:

ddutt:
One more comment in the code for FsCmd.py

suzieq/cli/sqcmds/FsCmd.py
```diff
@@ -0,0 +1,57 @@
...
+        # Get the default display field names
+        now = time.time()
+        if self.columns != ["default"]:
+            self.ctxt.sort_fields = None
+        else:
+            self.ctxt.sort_fields = []
+
+        df = self.sqobj.get(
+            hostname=self.hostname,
+            device=device,
```
ddutt, line 51: Make this device.split(). Same for mountPoint, but not for used-percent

suzieq/cli/sqcmds/FsCmd.py
```diff
@@ -0,0 +1,57 @@
...
+            view=view,
+            namespace=namespace,
+            columns=columns,
+            format=format,
+            sqobj=FsObj,
+        )
+
+    @command("show")
+    @argument("device", description="The Filesystem name")
+    def show(self, device: str = ''):
```
ddutt, line 35: Make device as list, not str. This is a bug on my part in how I did it in MacCmd.py. Do the same for mountPoint, but not for used-percent.

Its a list because we allow the user to specify more than one value

anubisg1:
i performed all changes required but it will need more work / specifically:

tests need to be reworked, as i have no ability to edit 
query_str in utils.py is bugged and specifically this is causing:

1) used_percent works only with "="
2) array built with typing.List[str] doesn't split items correctly

all of those are known to Dinesh and he agreed to fix them

PR #236, opened 2020-08-18
Modify bgp.yml to better reflect FRR

donaldsharp:
On the linux option to query FRR the command is
vtysh -c "show ip bgp ..."

This is deprecated syntax.  Modify to use the
non-deprecated syntax.

Signed-off-by: Donald Sharp <donaldsharp72@gmail.com>


donaldsharp:
yep, you're right.  fixed

ddutt:
Tested that "show bgp vrf all neighbors json" produces valid JSON output as well by running the output through json_pp

PR #420, opened 2021-09-16
Small documentation improvements

LucaNicosia:
The major modification is at `docs/getting_started.md:21-23` where I deleted a sentence which was referring to something called **mydatacenter** but never referenced in other md files.

### DIFF
--- a/docs/getting_started.md
+++ b/docs/getting_started.md
@@ -4,21 +4,23 @@ We want to make it as easy as possible for you to start engaging with Suzieq. We
 
 ### As a Docker Container
 
-To get started seeing the kind of analysis Suzieq provides today, we have a demo container that has data in including the the iamge.
-
+To get started seeing the kind of analysis Suzieq provides today, we have a demo container that has data in including the the image.
+```
     docker run -it -p 8501:8501 --name suzieq netenglabs/suzieq-demo
-    suzieq-cli for the CLI OR
-    suzieq-gui for the GUI. Connect to http://localhost:8501 via the browser to access the GUI
-
+    suzieq-cli // for the CLI
+    suzieq-gui // for the GUI. Connect to http://localhost:8501 via the browser to access the GUI
+```
 When you're within the suzieq-cli, you can run device unique columns=namespace to see the list of different scenarios, we've gathered data for. Use help inside the suzieq-cli to navigate your way around the CLI. Network operators should find the CLI easy to use as a router/bridge CLI as it provides contextual help and completions.
 
 To start collecting data for your network, create an inventory file to gather the data from following the instructions here. Decide the directory where the data will be stored (ensure you have sufficient available space if you're going to be running the poller, say 100 MB at least). Lets call this dbdir. Now launch the suzieq docker container as follows:
 ```
     docker run -it -vdbdir:/suzieq/parquet --name sq-poller netenglabs/suzieq
-    Connect to the container via docker attach sq-poller
+    docker attach sq-poller // connect to the container
 ```
 
-Launch the poller with the appropriate options. For example, sq-poller -D inventory.yml -k where mydatacenter is the name of the namespace where the data associated with the inventory is storedand inventory.yml is the inventory file in Suzieq poller native format (Use -a if you're using Ansible inventory file format).
+<!-- I think that the commented sentence below is what remains from an older documentation version -->
+
+Launch the poller with the appropriate options. For example, `sq-poller -D inventory.yml -k` where <!-- mydatacenter is the name of the namespace where the data associated with the inventory is stored and --> inventory.yml is the inventory file in Suzieq poller native format (Use `-a` if you're using Ansible inventory file format).
 
 ### As a Python Package
 
--- a/docs/getting_started.md
+++ b/docs/getting_started.md
@@ -4,18 +4,18 @@ We want to make it as easy as possible for you to start engaging with Suzieq. We
 
 ### As a Docker Container
 
-To get started seeing the kind of analysis Suzieq provides today, we have a demo container that has data in including the the image.
+To get started seeing the kind of analysis Suzieq provides today, we have a demo container that has data in including the image.
 ```
     docker run -it -p 8501:8501 --name suzieq netenglabs/suzieq-demo
-    suzieq-cli // for the CLI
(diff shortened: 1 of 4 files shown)


docs/getting_started.md
```diff
@@ -4,21 +4,23 @@ We want to make it as easy as possible for you to start engaging with Suzieq. We
 
 ### As a Docker Container
 
-To get started seeing the kind of analysis Suzieq provides today, we have a demo container that has data in including the the iamge.
-
+To get started seeing the kind of analysis Suzieq provides today, we have a demo container that has data in including the the image.
```
ddutt, line 7: Can you remove the second "the" please i..e change "including the the image" to "including the image"?

docs/getting_started.md
```diff
@@ -4,21 +4,23 @@ We want to make it as easy as possible for you to start engaging with Suzieq. We
...
+```
     docker run -it -p 8501:8501 --name suzieq netenglabs/suzieq-demo
```
ddutt, line 9: If you look at the original and this, this is broken. It looks like a comment in C++

docs/getting_started.md
```diff
@@ -4,21 +4,23 @@ We want to make it as easy as possible for you to start engaging with Suzieq. We
...
+    suzieq-cli // for the CLI
+    suzieq-gui // for the GUI. Connect to http://localhost:8501 via the browser to access the GUI
+```
 When you're within the suzieq-cli, you can run device unique columns=namespace to see the list of different scenarios, we've gathered data for. Use help inside the suzieq-cli to navigate your way around the CLI. Network operators should find the CLI easy to use as a router/bridge CLI as it provides contextual help and completions.
 
 To start collecting data for your network, create an inventory file to gather the data from following the instructions here. Decide the directory where the data will be stored (ensure you have sufficient available space if you're going to be running the poller, say 100 MB at least). Lets call this dbdir. Now launch the suzieq docker container as follows:
 ```
     docker run -it -vdbdir:/suzieq/parquet --name sq-poller netenglabs/suzieq
-    Connect to the container via docker attach sq-poller
+    docker attach sq-poller // connect to the container
```
ddutt, line 18: Change the // to # for commenting or better still put it in parentheses ()

docs/poller.md
```diff
@@ -31,13 +35,13 @@ The Suzieq native inventory file format that contains the IP address, the access
     - url: https://vagrant@192.168.123.123 password=vagrant
 ```
 
-There's a template in the docs directory called hosts-template.yml. You can copy that file as the template and fill out the values for namespace and url (remember to delete the empty URLs and to not use TABS, some editors add them automatically if the filename extension isn't right). The URL is the standard URL format: <transport>://[username:password]@<hostname or IP>:<port>. For example, ssh://dinesh:dinesh@myvx or ssh://dinesh:dinesh@172.1.1.23. 
+There's a template in the docs directory called hosts-template.yml. You can copy that file as the template and fill out the values for namespace and url (remember to delete the empty URLs and to not use TABS, some editors add them automatically if the filename extension isn't right). The URL is the standard URL format: `<transport>://[username:password]@<hostname or IP>:<port>`. For example, `ssh://dinesh:dinesh@myvx` or `ssh://dinesh:dinesh@172.1.1.23`.
```
ddutt, line 38: Do you think its worth adding a `` around the filename hosts-template.yml?

docs/poller.md
```diff
@@ -31,13 +35,13 @@ The Suzieq native inventory file format that contains the IP address, the access
...
 
 If you're using Ansible to configure the devices, an alternate to the native Suzieq inventory format is to use an Ansible inventory format. The file to be used is the output of the ```ansible-inventory --list``` command. 
 
 Once you have either generated the hosts file or are using the Ansible inventory file, you can launch the poller inside the docker container using **one** of the following two options: 
 
-* If you're using the native YAML hosts file, use the -D option like this: `sq-poller -D eos`  or
+* If you're using the native YAML hosts file, use the -D option like this: `sq-poller -D eos`
 * if you're using the Ansible inventory format, use the -a and -n options like this: via `sq-poller -a /suzieq/inventory -n eos`.
```
ddutt, line 45: Do you want to put `` around -a and -n like you did earlier for other options such as -v?

docs/poller.md
```diff
@@ -47,22 +51,22 @@ The poller creates a log file called /tmp/sq-poller.log. You can look at the fil
...
 
 * Put the password in the inventory file and ensure no one can read it
-* Use --ask-pass to then be prompted for a password
-* Use an environment var to store the password and pass the name of the env var via --envpass
+* Use `--ask-pass` to then be prompted for a password
+* Use an environment var to store the password and pass the name of the env var via `--envpass`
 * Use keyfile
-* Use passphrase protected keyfile  (use --passphrase)
+* Use passphrase protected keyfile  (use `--passphrase`)
 * Use ssh-config (not well-tested)
```
ddutt, line 58: While you're fixing this, could you please help me by removing the "(not well-tested)" line? Its been tested now.

docs/rest-server.md
```diff
@@ -7,20 +7,29 @@ Suzieq ships with a REST server. It has the same set of commands, verbs and filt
...
+The server is now accessible via [https://localhost:8000/docs](https://localhost:8000/docs) (or whatever port you've mapped the server to on the host). You need to pass the API_KEY in the request to be able to access the server. A simple example using the default API key and certificate is to use curl as follows:
+```
+    curl --insecure 'https://localhost:8000/api/v2/device/show?&access_token=496157e6e869ef7f3d6ecb24a6f6d847b224ee4f'
+```
 
-The Suzieq docker container of course serves the data available under /suzieq/parquet inside the container. You can mount the parquet data you've already gathered via the -v option. So an example of mounting the parquet directory with data would be to launch the container as follows:```docker run -itd -p 8000:8000 -v/home/netenglabs/work/suzieq/tests/data/multidc/parquet-out:/suzieq/parquet --name suzieq netenglabs/suzieq:latest```
+The Suzieq docker container of course serves the data available under /suzieq/parquet inside the container. You can mount the parquet data you've already gathered via the -v option. So an example of mounting the parquet directory with data would be to launch the container as follows:
```
ddutt, line 23: Add `` around -v please

ddutt (changes requested):
Please look at the comments and resolve them. Also, you need to sign-off on the commit. Use --signoff option with git commit to do so.

Issue #416, opened 2021-09-07
parser not matching OSPF External routes and default candidates


aegiacometti:
**Describe the bug**
Some routes are not being matched by the TextFSM template defined at file [iosxe_show_route.tfsm](https://github.com/netenglabs/suzieq/blob/master/suzieq/config/textfsm_templates/iosxe_show_route.tfsm), henced not being collected and showed with **suzieq-cli**.

**Examples:**
1.- OSPF "IA" routes
2.- candidate default routes

**Expected behavior**
Routes should appear in suzieq normally

**Additional context**
Running on GNS3 old IOS versions
3700 Software (C3725-ADVENTERPRISEK9-M), Version 12.4(15)T13, RELEASE SOFTWARE (fc3)

**To Reproduce**
Just execute the template [iosxe_show_route.tfsm](https://github.com/netenglabs/suzieq/blob/master/suzieq/config/textfsm_templates/iosxe_show_route.tfsm) with this file as input.
[example1.txt](https://github.com/netenglabs/suzieq/files/7124628/example1.txt)

You can use this simple script to do it
```
import textfsm
from tabulate import tabulate

template = sys.argv[1]
output_file = sys.argv[2]

with open(template) as f, open(output_file) as output:
    re_table = textfsm.TextFSM(f)
    header = re_table.header
    result = re_table.ParseText(output.read())
    print(result)
    print(tabulate(result, headers=header))
```

**Output example:**
```
vrf    protocol    prefix    preference    metric    nexthopIps    oifs                 statusChangeTimestamp    nexthopvrf
-----  ----------  --------  ------------  --------  ------------  -------------------  -----------------------  ------------
       O           1.0.1.0   110           41        ['3.3.0.2']   ['FastEthernet0/0']  00:01:07
       C           3.3.0.0                           []            ['FastEthernet0/0']
       O           3.0.0.0   110           40        ['3.3.0.2']   ['FastEthernet0/0']  00:01:07
       C           3.4.0.0                           []            ['FastEthernet0/1']
       O           20.0.0.0  110           41        ['3.3.0.2']   ['FastEthernet0/0']  00:01:09
       O           10.0.0.1  110           41        ['3.3.0.2']   ['FastEthernet0/0']  00:01:12
```

**Proposed solution:**
In the file [iosxe_show_route.tfsm](https://github.com/netenglabs/suzieq/blob/master/suzieq/config/textfsm_templates/iosxe_show_route.tfsm) replace the protocol value line
[...]

Issue #415, opened 2021-09-07
Parsing output "show ip route" not matching in iosxe_show_route.tfsm


aegiacometti:
**Describe the bug**
IOS "show ip route" command displays information differently when all routes share the same mask. This makes the parser fail to match the lines. Basically in suzieq the routes /24 are not being showed.

Check the 2 pictures as examples.
I'm marking the missing routes and some tips to suggest what to update in the parser file.
In example 1 all routes are /24. Follow the red word "subnetted", it is the key to solve this.
<img width="396" alt="example1" src="https://user-images.githubusercontent.com/16782779/132343727-238e02ea-dc4d-4373-93d1-f7e363af7544.png">

In example 2, I modified one of the routes to /25 and the output changed. All routes have the mask and check the key in red has changed to "variably subnetted"
<img width="399" alt="example2" src="https://user-images.githubusercontent.com/16782779/132343734-d389f208-af0c-486e-b075-9c725626c1a7.png">

**To Reproduce**
Just execute the template iosxe_show_route.tfsm with this new input. I'm attaching in text mode too.
[example1.txt](https://github.com/netenglabs/suzieq/files/7121443/example1.txt)
[example2.txt](https://github.com/netenglabs/suzieq/files/7121446/example2.txt)

 **Expected behavior**
Routes should appear in suzieq normally

**Additional context**
Running on GNS3 old IOS versions
3700 Software (C3725-ADVENTERPRISEK9-M), Version 12.4(15)T13, RELEASE SOFTWARE (fc3)

aegiacometti:
**For the issue 2:** regarding capturing OSPF **IA** routes, I see have two options to modify the template file **iosxe_show_route.tfsm**.

**Option1:** If capturing IA is important, then change:
Line `Value protocol ([A-Zi]\*?|[A-Zio]\s+[A-Z][1-2]|i ia|i su|LC)`
to `Value protocol ([A-Zi]\*?|[A-Zio]\s+[[A-Z]{2}]?\s|i ia|i su|LC)`
Since **[1-2]** is not matching the letter **A** from **IA**

**Option 2:** If we don't have to capture IA, then add a this line to have the match of **IA** lines
`^${protocol}\*?\s+[[A-Zi]{2}]?\s+${prefix}\s+\[${preference}/${metric}\]\s+via\s+${nexthopIps},\s+${statusChangeTimestamp},\s+${oifs}.*$$`

aegiacometti:
[...]

Issue #407, opened 2021-08-19
Add list of supported OS to Intro in read the docs


Neelajacques:
This will be one of the most common questions, let's pull same description from poler entry and put at the end of intro...

aegiacometti:
@Neelajacques please check the PR #424, one of the commits is related to this issue. If you are OK please close this issue.

Issue #406, opened 2021-08-19
Add IOS to Supported 


Neelajacques:
**Describe the bug**
List of supported OS is not up to date here: https://suzieq.readthedocs.io/en/latest/poller/#gathering-data

(missing IOS) 

Also, let's make alphabetical, consistent across sources... and do Vendor OS as a format 

From: "Currently Suzieq supports polling Cumulus Linux, Arista, Nexus, and Juniper and SONIC devices, as well as native Linux devices such as servers. Suzieq can easily support other device types, and we have third-party contributors working on other NOSes. Please let us know if you're interested in Suzieq supporting other NOSes."

Note extra And should also be removed

aegiacometti:
@Neelajacques please check the PR #424, one of the commits is related to this issue. If you are OK please close this issue.

PR #423, opened 2021-09-17
Fix IOS route parsing #415 #416

aegiacometti:
Fixing TextFSM template, adding some tests too. #415 and #416
Also some DOC typos and minor reading enhancements like #406 #407


aegiacometti:
fixing signoff in next PR

PR #426, opened 2021-09-27
Doc fixes

aegiacometti:
As I read the documentation, I propose some typos, format, reading, IOS mentions, etc


aegiacometti:
will fix singing and resend

PR #432, opened 2021-10-07
fix issue #431 - Missing BGP ASN number and incorrect prefixes assignment to AFI type

### LINKED ISSUE #431: Missing BGP ASN number and incorrect prefixes assignment to AFI type
**Describe the bug**
On Cisco IOS, for BGP there are missing ASN numbers and the prefixes are assigned to incorrect AFI types.
TextFSM templates are OK.

`poller/bgp.py` needs a minor fix in the looping of gathered data

**Screenshots**
Missing ASN numbers
issue_wrong_asn_and_prefixes_to_afi_type
Prefixes are being assigned to AFI VPNv4 when it should be IPv4
show_bgp_all_summary

aegiacometti:
Small update in the loop for processing gathered data (**show ip bgp all neigh**) at `suzieq/poller/services/bgp.py`

Previously the loop overwrites the last entry of (**show ip bgp neigh**) to add ASN and prefixes from (**show ip bgp all summ**), without taking into consideration adding to the correct AFY type.

By adding an extra KEY per AFI type, we can add the ASN to all the AFI types, and the prefixes only to the corresponding AFI type.

Now **suzieq-cli** displays the correct information.

fixed_wrong_asn_and_prefixes_to_afi_type

Pytest (as expected) returns the same output after this change.

Closes #431


aegiacometti:
sorry... signoff error again

PR #442, opened 2021-10-10
Add pylint linter to pre-commit

zxiiro:
This has most of the code files excluded from pylint since many files
currently do not pass but sets up the possibility for the project to
start updating files to conform to pylint. The excludes list can be
updated as this happens to start running pylint checks on said files.

Also adds pydoc strings to a few files missing them and resolve
pylint's W0107 error.

Ref: https://vald-phoenix.github.io/pylint-errors/plerr/errors/basic/W0107.html
Signed-off-by: Thanh Ha <zxiiro@gmail.com>


.pre-commit-config.yaml
```diff
@@ -8,3 +8,23 @@ repos:
     hooks:
       - id: end-of-file-fixer
       - id: mixed-line-ending
+
+  - repo: https://github.com/PyCQA/pylint
+    rev: v2.11.1
+    hooks:
+      - id: pylint
```
ddutt, line 15: I've been using flake8, and have found it generally better than pylint. Would you mind changing this to flake8? I want to merge it ASAP, and into 0.16.0 branch.

ddutt (changes requested):

.pre-commit-config.yaml
zxiiro: I think pylint and flake8 complement each other so I created a separate PR to add flake8 via #484 we can merge which ever one is more important first and then the other one after.

PR #448, opened 2021-10-11
Let poetry automatically pull version from tags

zxiiro:
Rather than spending time updating the 2 files that have hardcoded
versions in them. Let poetry auto-generate the versions directly from
Git tags.

If a dev has a git tag checked out then `poetry build` will generate
a version directly from that tag, for example the tag "v0.15.1" will
produce a version "0.15.1".

Otherwise the version will be based on how many commits since the last
tag was available in Git history for example "0.15.1-post.13+d211ed4347"
where "post.13" represents 13 commits since v0.15.1 was tagged and
"d211ed4347" is the commit hash of the commit that generated this
version.

For this to work however devs are required to install
poetry-dynamic-versioning in addition to poetry at the global level.

Ref: https://pypi.org/project/poetry-dynamic-versioning/
Signed-off-by: Thanh Ha <zxiiro@gmail.com>


zxiiro:
> @zxiiro Thank you for all the contributions. I was in the middle of a release and so didn't start to work on your new PRs. Will do so once the release is out, which should be early today

No worries. I'm just doing this for fun. Let me know if there's anything you need me to change in my PRs :)

PR #460, opened 2021-11-04
Add default columns to unique

### LINKED ISSUE #290: have unique default to columns
**Is your feature request related to a problem? Please describe.**
rather than just complain about not having a column default to hostname

claudious96:
This PR fixes #290 adding defaults columns to the `unique` command, allowing the user to simply run:
```
<any command here> unique
```
and getting a default column instead being forced to specify a column. i.e.:
```
ospf unique columns=hostname
```


suzieq/sqobjects/basicobj.py
```diff
@@ -223,7 +224,7 @@ def unique(self, **kwargs) -> pd.DataFrame:
         columns = kwargs.pop('columns', self.columns)
 
         if columns is None or columns == ['default']:
-            raise ValueError('Must specify columns with unique')
+            columns = self.default_columns
```
ddutt, line 227: User can specify '*' which is all columns instead of nothing or 'default', and we need to reject that as well. The original code had this bug too, just noticing it.

suzieq/sqobjects/basicobj.py
```diff
@@ -72,6 +72,7 @@ def __init__(self, engine_name: str = 'pandas',
             self.view = view or 'latest'
 
         self.columns = columns
+        self.default_columns = ['hostname']
```
ddutt, line 75: is default_columns the right name? That'd gibe another contributor the impression that this is used for all commands. Is _unique_def_column a better name? We also start var names that we use for internal use only with _, and so is _unique_def_column a better name?

ddutt (changes requested):

ddutt:
Claudio, you'll need to sign off on the commits with a --signoff option during the commit.

suzieq/sqobjects/basicobj.py
claudious96: Ok, I'll add that check too.

suzieq/sqobjects/basicobj.py
claudious96: You're right, it could be confusing to other contributors. It's better to go with `_unique_def_column`.

suzieq/sqobjects/path.py
```diff
@@ -7,6 +7,7 @@ def __init__(self, **kwargs):
         super().__init__(table='path', **kwargs)
         self._valid_get_args = ['namespace', 'hostname', 'columns',
                                 'vrf', 'source', 'dest']
+        self._unique_def_column = ['hopcount']
```
ddutt, line 10: I think hostname makes more sense here than hopCount. The reason is that hopCount is just an increasing number of hops from src to dest. Its not very helpful to get a list of 1, 2, 3 etc. each with how many ever paths there are. So, it'll be something like:
```
0 8
1 8
2 8
```
To indicate that there are 3 hopCounts (0, 1, 2), each with a count of 8

suzieq/sqobjects/device.py
```diff
@@ -9,6 +9,7 @@ def __init__(self, **kwargs):
         self._valid_get_args = ['namespace', 'hostname', 'columns', 'os',
                                 'vendor', 'model', 'status', 'version',
                                 'query_str']
+        self._unique_def_column = ['model']
```
ddutt, line 12: I'm torn between model and os here. Lets leave it at what you have for now.

ddutt (changes requested):

suzieq/sqobjects/device.py
claudious96: Yes I thought about os too but using os we'd get only the OS name without the version, i.e.:
```
          os
0    cumulus
1        eos
2        ios
3  junos-qfx
4      linux
5       nxos
```
If we then want to use os, I think it would be more useful to get both os and (os) version, i.e.:
```
          os
0    cumulus 3.7.9
1    cumulus 4.3.0
2        eos 4.23.8M
...
```
I'll leave `model` for the moment.

claudious96:
Using the assert modified with PR #466 all tests passes. I also fixed a few more things in [aaaaeff](https://github.com/netenglabs/suzieq/pull/460/commits/aaaaeff5b4c1c0bf3e7211ae9225a3e40cd37508).

PR #461, opened 2021-11-04
Issue 294

LucaNicosia:
This is a first version for the resolution of issue 294


suzieq/poller/services/interfaces.py
```diff
@@ -232,7 +232,7 @@ def _clean_cumulus_data(self, processed_data, raw_data):
                 first_entry.update({"master": entry["master"]})
 
             if entry.get('type', '') == "vxlan":
-                entry['speed'] = 0
+                entry['speed'] = -1
```
ddutt, line 235: Vxlan is a virtual interface and its speed SHOULD be 0, not -1.

suzieq/poller/services/interfaces.py
```diff
@@ -316,7 +316,7 @@ def _clean_junos_data(self, processed_data, raw_data):
                 elif entry['speed'].endswith('Gbps'):
                     entry['speed'] = int(entry['speed'].split('Gb')[0])*1000
                 elif entry['speed'] == 'Unlimited':
-                    entry['speed'] = 0
+                    entry['speed'] = -1
```
ddutt, line 319: This is another case where the speed should be 0. These interfaces are loopback interfaces, and are not included in interface assert check.

ddutt (changes requested):

LucaNicosia:
[outputcomment.log](https://github.com/netenglabs/suzieq/files/7487410/outputcomment.log)

I attached a yaml file (git doesn't allow me to push it so I renamed it to '.log') with some results of my job.

Please take a look at 'junos' and 'nxos' implementations and even a closer look at "Unlimited" or "auto" speed-value cases for these device types.

Now I start working on tests.

ddutt (changes requested):

suzieq/poller/services/interfaces.py
```diff
@@ -31,6 +30,25 @@ def _assign_vrf(self, entry, entry_dict):
...
                         entry_dict[intf] = {'vrf': entry['ifname']}
+    
+    @staticmethod
+    def get_missing_speed_value() -> int:
+        """Return the value for speed not retrieved"""
+        return -1
+
+    def _speed_field_check(self, entry, missing_speed_indicator):
+        """Check if the speed has valid value or not"""
+        if entry.get('speed',missing_speed_indicator) == missing_speed_indicator:
```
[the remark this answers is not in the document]
LucaNicosia: I think that this function should not be modified.
Take the firewall as an example: there will be a function called `_clean_firewall_data`. Suppose that the firewall replies with `None` if there is no speed. Now in the  `_clean_firewall_data`, we only have to call `_speed_field_check(entry,None)`. Another solution is to check in the  `_clean_firewall_data` body if the speed == None and change it to a -1 and then call `_speed_field_check(entry,-1)` or even directly don't call it.

PR #466, opened 2021-11-09
Fix assert_df_equal for list columns

claudious96:
This PR fixes a problem that arises in the test suite when comparing dataframes. The function `assert_df_equal` checks if two dataframes are equivalent trying to order a dataframe by columns and then using `pandas.Dataframe.compare`.
If one the columns contains that contains lists an exception is raised. Since lists are not hashable, pandas cannot perform the sort and therefore breaks, currently the tests try to avoid to order by certain columns that are known to contain lists. 

However in #460 a new column `ipAddress` is introduced in the `address` table and contains lists of ip addresses. The column is already whitelisted since in the `arpnd` table a column named `ipAddress` exists and contains only a string (1 ip address) and therefore can be used to sort.

To fix this, before sorting the dataframes, each value of each column is converted to a string. If a column contains lists as values, each element of each list is collapsed into a string. This way each column can be used for ordering.

Execution time of tests on github actions seems to be the same as before (~20 mins)

### DIFF
--- a/tests/integration/utils.py
+++ b/tests/integration/utils.py
@@ -64,30 +64,23 @@ def assert_df_equal(expected_df, got_df, ignore_cols) -> None:
             expected_df = expected_df.drop(
                 columns=ignore_cols, errors='ignore')
 
-    try:
-        if isinstance(got_df.index, pd.RangeIndex):
-            expected_df = expected_df \
-                .sort_values(by=expected_df.columns.tolist()) \
-                .reset_index(drop=True)
-        else:
-            expected_df = expected_df \
-                .sort_values(by=expected_df.columns.tolist())
+    # Detect which columns contain lists and convert lists to string
+    expected_df = expected_df.transform(_list_columns_to_str)
+    got_df = got_df.transform(_list_columns_to_str)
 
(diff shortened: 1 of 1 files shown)


tests/integration/utils.py
```diff
@@ -64,30 +64,23 @@ def assert_df_equal(expected_df, got_df, ignore_cols) -> None:
...
+            .reset_index(drop=True)
+    else:
+        expected_df = expected_df \
+            .sort_values(by=expected_df.columns.tolist())
+
+    if isinstance(expected_df.index, pd.RangeIndex):
+        got_df = got_df.sort_values(by=got_df.columns.tolist()) \
+                        .reset_index(drop=True)
+    else:
+        got_df = got_df.sort_values(by=got_df.columns.tolist())
```
ddutt, line 83: Don't you have other reasons why we need try/except?

claudious96: The only reason I encountered was that columns were not sortable. With this patch every column will become sortable.
I'm not saying there is no need at all for a try except, but since the current/old code only manages the unsortable column issue this patch will do the same.

PR #481, opened 2021-11-19
Add services validation at poller startup

claudiolor:
At the moment, when users specify the services to execute in the poller with the `-s` flag, or the ones to exclude with the `-x` flag, if one of them is not valid, they are silently ignored. This, for example, could cause some problems with typos:
If a user wants to exclude the `interfaces` services, they could launch:
```sq-poller -D inventory.yml -x interface```
instead of
```sq-poller -D inventory.yml -x interfaces```
Since no error is reported, they won't be aware of the error, until they realises that interfaces are actually polled. 

This PR solves this problem introducing validation and a clear error message for all the services specified with `-s` or `-x` flag.

### DIFF
--- a/suzieq/poller/sq_poller.py
+++ b/suzieq/poller/sq_poller.py
@@ -206,20 +206,36 @@ async def start_poller(userargs, cfg):
 
     ignore_known_hosts = userargs.ignore_known_hosts or None
 
+    # Retrieve the services to start
+    svcs = list(Path(cfg["service-directory"]).glob('*.yml'))
+    allsvcs = [os.path.basename(x).split('.')[0] for x in svcs]
+    svclist = None
+
     if userargs.service_only:
         svclist = userargs.service_only.split()
+
+        # Check if all the given services are valid
+        notvalid = [s for s in svclist if s not in allsvcs]
+        if notvalid:
+            print(f'Invalid sevices specified: {notvalid}. '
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Review by LucaNicosia — APPROVED** [review]

**Review by claudious96 — APPROVED** [review]

File: suzieq/poller/sq_poller.py [hunk 1]
```diff
@@ -206,20 +206,36 @@ async def start_poller(userargs, cfg):
...
-        svclist = [os.path.basename(x).split('.')[0] for x in svcs]
+        svclist = allsvcs
 
     if userargs.exclude_services:
-        svclist = list(filter(lambda x: x not in userargs.exclude_services,
+        excluded_services = userargs.exclude_services.split()
+        # Check if all the excluded services are valid
+        notvalid = [e for e in excluded_services if e not in allsvcs]
+        if notvalid:
+            print(f'Services {notvalid} excluded, but they '
```
ddutt, line 231: Why should we continue in this case? For example, if the user typed iterfaces and we continued, that'd be not good, especially if we're not doing so in the case of the other option, "-s". I think you should terminate this as well.

ddutt (changes requested):

ddutt:
I'll add a couple of tests to verify these checks work correctly. I'll do them this time. Please add them the next time.

PR #485, opened 2021-11-23
Run pre-commit as part of PR checks

zxiiro:


zxiiro:
I fully expect this to fail due to existing flake8 issues as well as some file line endings issues. Once the flake8 issues have been resolved I'll rebase and let pre-commit auto resolve the line endings and we should be good to have this enabled moving forward.

PR #492, opened 2021-11-25
Fix #490

### LINKED ISSUE #490: Poller Parser Errors for Junos MX
**Describe the bug**

In parsing Junos MX devices, the following errors were found:

- BGP's bfdStatus had a state of initializing instead of ['up', 'down', 'disabled']
- VPLS interfaces had an incorrect state/adminState values
- Route parsing MX routes to exclude BGP routes which might be too big on MX causes exceptions in route parsing
- Some MAC entries have a vlan value of "none" which results in an exception as the code tries to convert this to int
- In some situations, there was no forwarding action specified, and so we have a blank instead of forward or blackhole

ddutt:


claudious96:
@ddutt in #490 is mentioned only Junos but in this PR there are fixes to iosxr clean up functions too, is that correct?

ddutt:
> @ddutt in #490 is mentioned only Junos but in this PR there are fixes to iosxr clean up functions too, is that correct?

Yes, you're correct. I found an IOSXR issue in the process and added that too as it was in the same file.

Issue #408, opened 2021-08-23
Add Palo Alto Networks panOS support


gpipito:
**Is your feature request related to a problem? Please describe.**
It would be good to add palo alto support to suzieq

**Additional context**
PanOS supports REST calls
https://docs.paloaltonetworks.com/pan-os/9-0/pan-os-panorama-api/get-started-with-the-pan-os-rest-api/pan-os-rest-api.html

ddutt:
Fixed, available since version 0.16.0. Try the latest version please.

PR #493, opened 2021-11-25
Panos support

claudious96:
This PR brings an initial support for PAN-OS Firewalls  (#408) . The following services are going to be supported in decreasing order of priority:

- [x] device
- [x] interfaces
- [x] routes (partial, route distinguisher is missing)
- [x] lldp
- [x] arpnd
- [x] bgp
- [ ] ospf
- [ ] vlan
- [ ] inventory
- [ ] mac


suzieq/poller/worker/nodes/node.py
```diff
@@ -1434,3 +1444,231 @@ async def rest_gather(self, service_callback, cmd_list, cb_token,
...
+
+    def _init(self, **kwargs):
+        super()._init(**kwargs)
+        self.devtype = "panos"
+
+    async def init_node(self):
+        try:
+            res = []
+            # temporary hack to detect device info using ssh
+            async with asyncssh.connect(
```
claudiolor, line 1459: Why do you need to create a separate process to connect to the device?

claudiolor (approved):

PR #494, opened 2021-11-26
sq_poller.py refactoring

claudiolor:
This PR contains a refactoring of the poller code:
- Introducing the poller class
- Introducing a plugin system for the output format
- Introducing a plugin system for the inventory source
- Clean and split the sq_poller.py code in smaller files


suzieq/poller/coalescer.py
```diff
@@ -0,0 +1,93 @@
+import errno
```
ddutt, line 1: I don't quite like the name coalescer.py because it gives the wrong impression that this is the real coalescer. Its small enough to put it in the poller?  Not completely against it, i.e. you can convince me to pursue your approach, but I need a little convincing.

Should we move sq_coalescer.py to this directory?

ddutt (changes requested):

PR #495, opened 2021-11-29
Simnode refactoring

claudious96:
This PR refactors the sim node related files in order to prepare to use it in the future for testing.

The simnode now doesn't need a list of available commands hardcoded, as long as the files in nossim follow the guidelines:
command: `show mac-address table` -> filename: `show_mac_address_table.<format>`
In case a `*` is used it is replaced with `all` in the filename.
At the moment there is one exception that cannot be automatically retrieved and there will be more in the future but the majority of the filename can be deduced from the command.

The simnode will be run as:
`tests/utilities/simnode_ssh.py --nos panos [--nos-version 8.0.0 --hostname leaf01 --listening-port 10000]`
Where options in square brackets are optional. 

In addition, it brings a `simnode_rest ` that provides XML output only for panos. It can be extended to provide also JSON REST APIs.

It also introduces `symnode.py` that, given a yaml file that describes the devices, starts the related processes to simulate the specified devices.


ddutt (approved):

ddutt (approved):
Please verify there are no linting errors

ddutt (approved):

ddutt (changes requested):
Can you split the addition to pyproject.toml and the corresponding poetry.lock and build/requirements.txt from this commit?

LucaNicosia (approved):

claudiolor (approved):

tests/utilities/simnode.py
```diff
@@ -0,0 +1,94 @@
...
+                "required": [
+                    "nos", "version", "hostname"
+                ]
+            }
+        }
+    }
+}
+
+
+def run_simnodes(devices: list):
```
ddutt, line 40: Sorry for being pedantic, but can you please add docstrings?

claudious96: No problems, should be done now

PR #497, opened 2021-11-29
Add SqPlugin base class

claudiolor:
This PR introduces a base class for all the Suzieq plugins, allowing to uniform the plugins discovery logic for every kind of plugin in the entire codebase.

### DIFF
--- a/suzieq/shared/sq_plugin.py
+++ b/suzieq/shared/sq_plugin.py
@@ -1,0 +1,70 @@
+"""
+This module contains all the common logic of
+the Suzieq plugins
+"""
+
+from importlib import import_module
+from importlib.util import find_spec
+from inspect import getmembers, getmro, isclass
+from pkgutil import iter_modules
+from typing import Dict, Type
+
+
+# pylint: disable=too-few-public-methods
+class SqPlugin:
+    """SqPlugin is the base common class inherited by all the
+    Suzieq plugins
+    """
+
+    def __init__(self) -> None:
+        pass
+
+    @classmethod
+    def get_plugins(cls,
+                    search_pkg: str,
+                    transitive=False) -> Dict[str, Type]:
+        """Discover all the plugins in the search_pkg package, inheriting
+        the current base class.
+
+        Args:
+            search_pkg (str): The package where to look for the plugins.
+                Note that if the specified package contains other packages,
+                the function will scan only their content, without accessing
+                to other nested packages and ignoring all the files inside
+                search_pkg.
+            transitive (bool, optional): If False returns a plugin only if
+                it is a direct descendant of the base class.
+                Defaults to False.
+
+        Returns:
+            Dict[str, Type]: a dictionary containing all
+                the descendants of the base class in the search_pkg,
+                with format {'module name': PluginClass}
+        """
+        classes = {}
+
+        # Collect the list of packages where to search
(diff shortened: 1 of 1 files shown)


suzieq/shared/sqPlugin.py
```diff
@@ -0,0 +1,66 @@
...
+                    mname = f'{pkg}.{use_name}'
+                    m = import_module(mname)
+                    for mbr in getmembers(m, isclass):
+                        if (mbr[1].__module__ == mname
+                           and (transitive or getmro(mbr[1])[1] == cls)
+                           and issubclass(mbr[1], cls)
+                           and mbr[1] != cls):
+                            classes[use_name] = mbr[1]
+
+        return classes
```
ddutt, line 66: Do we need an abstract property or method called "plugin_type" that identifies the plugin type?

ddutt (changes requested):
Please review the linting errors

suzieq/shared/sqPlugin.py
```diff
@@ -0,0 +1,66 @@
...
+    """SqPlugin is the base common class inherited by all the
+    Suzieq plugins
+    """
+
+    @abc.abstractmethod
+    def __init__(self) -> None:
+        pass
+
+    @classmethod
+    def discover_plugins(cls,
```
ddutt, line 19: I don't like the name discover_plugins that returns a dict of plugins. Is a name such as get_plugins better OR we can have discover_plugins which updates a class variable called plugins?

suzieq/shared/sqPlugin.py
claudiolor: As we discover by the class I was thinking about a solution using the Class type to identify a plugin type. I'm sure having a property allows to easily know the type of the plugin once you have one of it, but at point shouldn't we have a function allowing to discover the plugin by the value of the property "type"?

ddutt: Thats interesting. Can you propose some code for this?

claudiolor: The code does not change so much, I think it should be enough declaring an class attribute like:
```
class SqPlugin:
         plugin_type = "BasePlugin"
         ...
```
Overriding this attribute every time we inherit SqPlugin from a plugin base class (i.e. SqObject).
From the point of view of the the `get_plugin()` method, we don't look anymore to the class but on the plugin_type string. I.e.
```
...
for mbr in getmembers(mod, isclass):
    if (mbr[1].__module__ == mname
        mbr[1].plugin_type == cls.plugin_type
        and mbr[1] != cls):
        classes[use_name] = mbr[1]
...
``` 
At this point, this solution allows to easily retrieve the kind of plugin by getting the plugin_type class attribute, but I think we might create some problems than the ones we solve:

- We need to manually set the `plugin_type` on the plugin base classes (i.e. in the SqObjects class), but it might be prone to errors (plugin_type not overwritten, typos, "service" instead of "Service"...)
- Classes having different parents but with the same `plugin_type` attribute are considered the same type of plugin, is it logically correct?

suzieq/shared/sqPlugin.py
claudiolor: I think they are both equally good, Probably, only renaming to `get_plugins()` should be enough, since plugin discovery is typically a one-shot operation we might not need to store the result. However, if you prefer the class variable is completely fine for me

PR #500, opened 2021-12-01
SqPlugin: Make get_plugins work without any additional params for a p…

ddutt:
…lugin

SqPlugin has a class method called get_plugins that returns all the
plugins associated with a plugin type. However this required an additional
parameter called search_pkg which was not really necessary. It can be
derived automatically from the module name. This patch does that.

Additionally, the name of the plugin returned when the plugin was in
a subdir as opposed to the same dir was not correct. This patch fixes
that as well.

The method signature isn't changed, and so if a caller wants to specify
the search path that still works

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -1,0 +1,872 @@
+[MASTER]
+
+# A comma-separated list of package or module names from where C extensions may
+# be loaded. Extensions are loading into the active Python interpreter and may
+# run arbitrary code.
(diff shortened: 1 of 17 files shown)


claudiolor (approved):

LucaNicosia (commented):
The code is ok for me. I ask you to add the file `__init__.py` in `suzieq/shared` before pushing

PR #503, opened 2021-12-04
Remote ux

ddutt:
This adds support for allowing the CLI, GUI and python API to run remotely from where the data is stored. A common ask has been to separate the CLI from where the data is stored because the poller maybe running on a privileged machine. 

I haven't added much security and the API key is still coming from the CLI instead of from an environment var. We'll discuss all this in the office hours.

### DIFF
--- a/suzieq/cli/sqcmds/command.py
+++ b/suzieq/cli/sqcmds/command.py
@@ -15,10 +15,11 @@ from colorama import Fore, Style
 
 from suzieq.shared.sq_plugin import SqPlugin
 from suzieq.shared.exceptions import UserQueryError
+from suzieq.shared.utils import SUPPORTED_ENGINES
 
 
 @dataclass
-class ArgHelpClass(object):
+class ArgHelpClass:
     '''Class holding description and other params to display help
     This class exists to simplify displaying help for vars that we
     cannot access via the classargspec var
@@ -29,7 +30,7 @@ class ArgHelpClass(object):
 @argument(
     "engine",
     description="Which analytical engine to use",
-    choices=["rest", "pandas"],
+    choices=SUPPORTED_ENGINES,
 )
 @argument(
     "namespace", description="Space separated list of namespaces to qualify"
@@ -63,7 +64,7 @@ class SqCommand(SqPlugin):
 
     def __init__(
             self,
-            engine: str = "pandas",
+            engine: str = "",
             hostname: str = "",
             start_time: str = "",
             end_time: str = "",
@@ -74,7 +75,7 @@ class SqCommand(SqPlugin):
             query_str: str = " ",
             sqobj=None,
     ) -> None:
-        self.ctxt = context.get_context()
+        self.ctxt = context.get_context().ctxt
         self._cfg = self.ctxt.cfg
         self._schemas = self.ctxt.schemas
         self.format = format or "text"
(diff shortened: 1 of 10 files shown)


suzieq/cli/sqcmds/context_commands.py
```diff
@@ -28,15 +28,32 @@
...
+    'rest_server_port', description='Port of the REST server'
+)
+@argument(
+    'rest_api_key', description='API key for the REST server'
+)
+@argument(
+    'rest_use_https', description='Use HTTPS for the REST server',
+    choices=['True', 'False']
+)
 def set_ctxt(
```
claudiolor, line 44: We need to change the description for the engine argument

suzieq/cli/sqcmds/context_commands.py
```diff
@@ -28,15 +28,32 @@
...
         engine: str = "",
         datadir: str = "",
         col_width: int = 50,
+        rest_server_ip: str = "",
+        rest_server_port: str = "",
+        rest_api_key: str = "",
+        rest_use_https: str = ""
 ):
     """set certain contexts for subsequent commands. Cmd is additive"""
     plugin_ctx = context.get_context()
```
claudiolor, line 59: You need to set the parameter in plugin_ctx.ctxt not in the NubiaSuzieqContext object

suzieq/shared/context.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+@dataclass
+class SqContext:
+    '''Context passed between the front-end and back-end'''
+    cfg: Dict = field(default_factory=dict)
+    schemas: Dict = field(default_factory=dict)
+
+    pager: bool = False
+    namespace: str = ''
```
claudiolor, line 12: It would be nice be consistent with single and double quotes

claudiolor (changes requested):

suzieq/shared/context.py
ddutt: Ack

suzieq/cli/sqcmds/context_commands.py
ddutt: Ack

suzieq/cli/sqcmds/context_commands.py
ddutt: Fixed

claudiolor (approved):

suzieq/engines/rest/engineobj.py
```diff
@@ -0,0 +1,89 @@
...
+            table = table[:-1]
+        url = (
+            f'{self.ctxt.rest_transport}://{self.ctxt.rest_server_ip}'
+            f':{self.ctxt.rest_server_port}'
+            '/api/v2/'
+            f'{table}/'
+            f'{verb}?'
+            f'{query_params}')
+
+        response = requests.get(url, verify=False, )
```
claudiolor, line 84:
```suggestion
        response = requests.get(url, verify=False)
```

claudiolor: Could be useful having an option in the CLI to enable and disable `verify`?

ddutt: Possibly. The thing is most companies I know don't deploy valid certificates. This is a question I have for the community during office hours. I deliberately didn't provide an option. I don't want to provide knobs that very few use, and they might have not used it too if it wasn't there. Maybe its more important to provide for that instead of http OR https.

suzieq/cli/sqcmds/command.py
```diff
@@ -29,7 +29,7 @@ class ArgHelpClass(object):
 @argument(
     "engine",
     description="Which analytical engine to use",
-    choices=["pandas"],
+    choices=["rest", "pandas"],
```
LucaNicosia, line 32: Why don't use SUPPORTED_ENGINES

claudious96 (approved):

LucaNicosia (approved):

PR #507, opened 2021-12-08
Poller minor fixes

claudiolor:
This PR contains some minor fixes on the poller:
- Ignore an empty service yml file without throwing an exception
- The no_coalescer user argument is not ignored now
- Have consistent name for 'user_password' both in Poller and ServiceManager objects
- Check if the Parquet output directory is not a file
- Add an attribute in the CoalescerLauncher for the coalescer process PID
- Improve Inventory documentation
- Refactor inventory dir structure
- Move poller into poller/worker


suzieq/poller/writers/parquet.py
```diff
@@ -45,7 +46,7 @@ def write_data(self, data: Dict):
         Args:
             data (Dict): dictionary containing the data to store.
         """
-        cdir = "{}/{}/".format(self.root_output_dir, data["topic"])
+        cdir = f"{self.root_output_dir}/{data['topic']}/"
```
LucaNicosia, line 49: Isn't it better to use `os.path.join` ?

suzieq/poller/writers/output_worker.py
```diff
@@ -15,19 +16,21 @@ class OutputWorker(SqPlugin):
...
                 os.makedirs(output_dir)
+            elif not os.path.isdir(output_dir):
+                raise SqPollerConfError(f'Output directory {output_dir}'
+                                        'is not a directory')
         else:
             # TBD: The right error to raise here since this is a required
             # keyword
-            self.logger.error("Need mandatory keyword arg: output_dir")
-            raise ValueError
+            raise SqPollerConfError('Need mandatory keyword arg: output_dir')
```
LucaNicosia, line 33: Is there a reason why you removed the log?
I think the comment right above should be removed

claudiolor: Log is generated at the exception catch level

claudious96 (approved):

LucaNicosia (approved):

suzieq/poller/inventory/inventory_sources_base/inventory.py
```diff
@@ -37,7 +37,7 @@ def __init__(self, add_task_fn: Callable, **kwargs) -> None:
         self.jump_host = kwargs.pop('jump_host', None)
         self.jump_host_key_file = kwargs.pop('jump_host_key_file', None)
         self.ignore_known_hosts = kwargs.pop('ignore_known_hosts', False)
-        self.user_password = kwargs.pop('password', None)
+        self.user_password = kwargs.pop('user_password', None)
```
ddutt: username and password are fairly well-known terms. Why rename it user_password? Just a q, doesn't affect the approval whichever way you proceed

suzieq/poller/writers/parquet.py
```diff
@@ -41,12 +44,12 @@ def __init__(self, **kwargs):
...
 
     def write_data(self, data: Dict):
-        """Write the data into the Parquet output directory"
+        """Write the data into the Parquet output directory
 
         Args:
             data (Dict): dictionary containing the data to store.
         """
-        cdir = f"{self.root_output_dir}/{data['topic']}/"
+        cdir = os.path.join(self.root_output_dir, data['topic'])
```
ddutt, line 52: Not that it matters, but f-strings are more performant than os.path.join I presume, and isn't python3 recommending the use of Pathlib? Again, just curious, does not affect approval whatever you choose.

suzieq/poller/worker/inventory/inventory_sources/ansible.py
```diff
@@ -5,7 +5,7 @@
 
 from typing import Dict, List
 
-from suzieq.poller.inventory.inventory_sources_base.inventory import Inventory
+from suzieq.poller.worker.inventory.inventory_sources_base.inventory import Inventory
```
ddutt: I forget now. Why is the inventory under worker and not under poller?

ddutt (approved):

PR #514, opened 2021-12-19
Allow all verbs to support all filters

ddutt:
Of course not help and describe.

This is a draft. Wanted you three to get some feedback to me sooner than later.


suzieq/cli/sqcmds/InventoryCmd.py
```diff
@@ -64,18 +56,12 @@ def show(self, type: str = "", status: str = "", model: str = "",
         if model:
             model = re.split(r"\s+(?=[^']*(?:'))", model)
         else:
-            model = []
+            vendor = []
```
claudious96, line 59: I think this is a typo, vendor is set as [] in the else block 4 lines above.

suzieq/cli/sqcmds/command.py
```diff
@@ -291,41 +326,57 @@ def _do_help(self, table: str, verb: str = 'show'):
...
-            if fnargs:
-                classargspec.update(fnargs)
-            classargspec.update(self._additional_help_vars)
-            for arg in sorted(classargspec):
-                print(f" - {arg}: " + Fore.CYAN +
-                      f"{classargspec[arg].description}" + Style.RESET_ALL)
-        else:
+        # Determine if the method is present
+        if verb == 'assert':
+            verb = 'aver'
```
claudious96, line 331: What happens here, why does it get changed to 'aver'?

suzieq/cli/sqcmds/command.py
```diff
@@ -203,11 +237,11 @@ def describe(self):
...
 
-    @command("top", help="find the top n values for a field")
-    @argument("count", description="number of rows to return")
-    @argument("what", description="integer field to get top values for")
-    @argument("reverse", description="return bottom n values",
-              choices=['True', 'False'])
+    @ command("top", help="find the top n values for a field")
+    @ argument("count", description="number of rows to return")
+    @ argument("what", description="numeric field to get top values for")
+    @ argument("reverse", description="return bottom n values",
```
LucaNicosia, line 243: Only visual comment. These spaces between `@` and the command shouldn't exists

suzieq/engines/pandas/address.py
```diff
@@ -80,14 +81,17 @@ def get(self, **kwargs) -> pd.DataFrame:
...
-        if 6 in addr_types:
-            df = df.explode('ip6AddressList').fillna({'ip6AddressList': ''})
+        if 6 in addr_types or ipvers in ["v6", ""]:
+            # df = df.explode('ip6AddressList').fillna({'ip6AddressList': ''})
+            addrcols.append('ip6AddressList')
 
-        if 'ipAddress' in columns or columns == ['*']:
-            ndf = pd.DataFrame(df[['ipAddressList', 'ip6AddressList']].agg(
+        if ('ipAddress' in columns or (columns == ['*'])) and not ipvers:
+            ndf = pd.DataFrame(df[addrcols].agg(
```
LucaNicosia, line 94: I think the result is going to be different from the older version since you are no more using the `df.explode` function. This result format modification was planned or not?

suzieq/cli/sqcmds/command.py
ddutt: Becaise the function is called aver, not assert. assert is a reserved keyword in python that cannot be used as a function name.

suzieq/cli/sqcmds/command.py
ddutt: I agree. That thing is added by autopep8 or something, I have to see which one. I don't add the space when I write the code

suzieq/engines/pandas/address.py
ddutt: Yes, that was intended. When I do a show, I didn't expect the addresses to be exploded, but they are now, after the prefix support was added. So, I fixed it.

PR #515, opened 2021-12-19
Fix handling of query-str with double quotes

### LINKED ISSUE #505: query_str must have a trailing whitespace on version 0.16.0
**Describe the bug**
If the query_str doesn't end with a whitespace, it will raise an error

**To Reproduce**
Doing the following command
```
interface show query-str="hostname=='spine02'" 
```
wiil result in:
```
ERROR: UserQueryError: EOL while scanning string literal (<unknown>, line 1)
``` 
The following command instead will work (note the whitespace between `'` and `"`)
```
interface show query-str="hostname=='spine02' " 
```

claudious96:
This PR fixed #505 adding a workaround for a known behavior of Nubia which strips the trailing single/double quote in case is the last character of the query string.

It also removes unused code in `engineobj` since double quotes are already stripped when the string gets there.

### DIFF
--- a/suzieq/cli/sqcmds/command.py
+++ b/suzieq/cli/sqcmds/command.py
@@ -87,7 +87,7 @@ class SqCommand(SqPlugin):
             # This happens because nubia strips off the trailing quote
             # if not followed by a blank
             query_str += '"'
-        if query_str.count("'") % 2 != 0:
+        elif query_str.count("'") % 2 != 0:
             # This happens because nubia strips off the trailing quote
             # if not followed by a blank
             query_str += "'"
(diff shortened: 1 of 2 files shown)


suzieq/cli/sqcmds/command.py
```diff
@@ -87,6 +87,10 @@ def __init__(
             # This happens because nubia strips off the trailing quote
             # if not followed by a blank
             query_str += '"'
+        if query_str.count("'") % 2 != 0:
```
claudiolor, line 90:
```suggestion
        elif query_str.count("'") % 2 != 0:
```
I think there is no need to to evaluate this expression if the previous one is true

claudiolor (approved):

LucaNicosia (approved):

ddutt:
I think the fix should be made only in the CLI part of the code since its a Nubia bug. Why do we need to fix it in the engineobj code?

claudious96:
@ddutt exactly. The bug is fixed in `command.py`, in `engineobj.py` I just removed unreachable lines of code. That is because when the string gets there it hasn't anymore the enclosing quotes and therefore the if condition is never verified.

PR #519, opened 2021-12-22
New poller fixes

LucaNicosia:
PR for 0.16.0 alpha release
- Fixing problems with -i command
- Wait for the coalescer before exiting the program
- Adding inventory validation for wrong fields
- Other small fixes

### DIFF
--- a/suzieq/config/etc/suzieq-cfg.yml
+++ b/suzieq/config/etc/suzieq-cfg.yml
@@ -29,9 +29,9 @@ poller:
   # logsize is in bytes
   # logsize: 10000000
   # log-stdout: True
-  # inventory-file: "suzieq/conf/etc/inventory/inventory.yaml"  # Inventory file to use if-I is not specified
+  # inventory-file: "suzieq/conf/etc/inventory/inventory.yaml"  # Inventory file to use if it is not specified
   # inventory-timeout: 10                                       # The maximum time in seconds for a source to
-  # return its node list to the poller
+                                                                # return its node list to the poller
   # update-period: 3600                                         # The update period of the inventory in seconds
 
   # The manager is the component of the poller producing the final inventory
@@ -40,8 +40,6 @@ poller:
   #
   # manager:
   #   workers: 1  # Number of workers polling the nodes
-  #   inventory-path: suzieq/.poller/intentory/static_inventory
-  #   inventory-file-name: static_inv
 
   # The chunker is the component of the poller in charge of splitting the
   # final inventory into a number of chunks equal to the number of workers
--- a/suzieq/poller/controller/controller.py
+++ b/suzieq/poller/controller/controller.py
@@ -275,7 +275,7 @@ class Controller:
         except asyncio.CancelledError:
             logger.warning('Received termination signal, terminating...')
 
-    def _stop(self):
+    async def _stop(self):
         """Stop the controller"""
 
         tasks = [t for t in asyncio.all_tasks()
--- a/suzieq/poller/controller/controller.py
+++ b/suzieq/poller/controller/controller.py
@@ -89,8 +89,8 @@ class Controller:
                     raise SqPollerConfError(
                         'Inventory file not found in the default location:'
                         f'{inventory_file}, use -I argument to provide it, '
-                        'use instead -i to provide an input directory '
-                        'and simulate an input.'
+                        'use -i instead to provide an input directory '
+                        'with pre-captured output and simulate an input.'
                     )
         else:
             if not Path(self._input_dir).is_dir():
--- a/suzieq/poller/controller/controller.py
+++ b/suzieq/poller/controller/controller.py
@@ -73,19 +73,30 @@ class Controller:
         self._validate_controller_args(args, config_data)
 
         # Get the inventory
-        inventory_file = args.inventory or self._config.get('inventory-file')\
-            or 'suzieq/conf/etc/inventory/inventory.yaml'
+        default_inventory_file = 'suzieq/config/etc/inventory.yaml'
+        inventory_file = None
 
-        if inventory_file:
+        if not self._input_dir:
+            inventory_file = args.inventory or \
+                             self._config.get('inventory-file') or \
+                             default_inventory_file
             if not Path(inventory_file).is_file():
+                if inventory_file != default_inventory_file:
+                    raise SqPollerConfError(
+                        f'Inventory file not found at {inventory_file}'
+                    )
+                else:
+                    raise SqPollerConfError(
+                        'Inventory file not found in the default location:'
+                        f'{inventory_file}, use -I argument to provide it, '
+                        'use instead -i to provide an input directory '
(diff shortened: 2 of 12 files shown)


suzieq/poller/controller/source/ansible.py
```diff
@@ -85,8 +85,9 @@ def _get_inventory(self) -> Dict:
                 password = entry['ansible_password']
 
             # Retrieve password information
-            devtype = entry.get('ansible_network_os', None)
-            if devtype == 'eos':
+            devtype = None
+            if entry.get('ansible_network_os') in ['eos', 'panos']:
+                devtype = entry.get('ansible_network_os')
```
ddutt, line 88: The correct fix is to check for https and only use devtype if the transport is https. The second thing is to allow device autodetection to override the devtype. This fix will have to be redone.

suzieq/poller/controller/controller.py
```diff
@@ -73,19 +73,30 @@ def __init__(self, args: argparse.Namespace, config_data: dict) -> None:
...
             if not Path(inventory_file).is_file():
+                if inventory_file != default_inventory_file:
+                    raise SqPollerConfError(
+                        f'Inventory file not found at {inventory_file}'
+                    )
+                else:
+                    raise SqPollerConfError(
+                        'Inventory file not found in the default location:'
+                        f'{inventory_file}, use -I argument to provide it, '
+                        'use instead -i to provide an input directory '
```
ddutt, line 92: Change the last sentence in the message to " Use -i instead to provide an input directory with pre-captured output"

suzieq/config/etc/suzieq-cfg.yml
```diff
@@ -40,8 +40,6 @@ poller:
   #
   # manager:
   #   workers: 1  # Number of workers polling the nodes
-  #   inventory-path: suzieq/.poller/intentory/static_inventory
-  #   inventory-file-name: static_inv
```
ddutt, line 44: Is this just removed from the example OR is this truly gone?

suzieq/poller/controller/credential_loader/base_credential_loader.py
```diff
@@ -35,6 +35,10 @@ def __init__(self, init_data: Dict) -> None:
         self._conf_passphrase = None
         self._conf_keyfile = None
         self._conf_username = None
+        self._valid_fields = ['username', 'name', 'type',
+                              'password', 'ssh-passphrase', 'keyfile']
+
+        self._validate_config(init_data)
```
ddutt, line 41: Do we need a mandatory fields check i.e. for an auth are some fields mandatory such as name and type and if its missing, complain and abort?

suzieq/poller/controller/source/base_source.py
```diff
@@ -26,8 +26,10 @@ def __init__(self, input_data) -> None:
         self._inv_is_set = False
         self._inv_is_set_event = asyncio.Event()
         self._name = input_data.get('name')
-        self._auth = input_data.get('auth')
+        self._auth = input_data.get('auth')  # auth object
         self._device = input_data.get('device') or {}
+        self._valid_fields = ['name', 'type', 'auth',
+                              'device', 'namespace', 'run_once']
```
ddutt, line 32: Same comment as in auth. DO we need a mandatory fields check?

suzieq/poller/controller/source/base_source.py
```diff
@@ -40,10 +42,10 @@ def __init__(self, input_data) -> None:
             'jump_host_key_file',
             'ignore_known_hosts'
         ]
-        errors = self._validate_config(input_data)
-        if errors:
-            raise InventorySourceError('Inventory validation failed: {}'
-                                       .format(errors))
+
+        self._validate_config(input_data)
+        self._validate_device()
```
ddutt, line 47: Do we also want to complain about keywords we don't know? This can happen because of a typo for example.

ddutt (changes requested):

suzieq/config/etc/suzieq-cfg.yml
claudiolor: We do not support these options anymore

LucaNicosia: This is truly gone. I forgot to delete this part from the configuration file while removing the parameter

suzieq/poller/controller/credential_loader/base_credential_loader.py
LucaNicosia: The mandatory field check is done in each different plugin. For the default auth credentials plugin, none of the fields is mandatory. For example we can use the native format in this way
```yaml
sources:
- name: native0
  hosts:
  - url: ssh://vagrant@1.1.1.1
  - url: ssh://vagrant@1.1.1.2
auths:
- name: auth0
  password: vagrant
namespaces:
- name: native-ns
  source: native0
  auth: auth0
```
If some mandatory fields are needed by a plugin, it is its duty to check them. For example check https://github.com/netenglabs/suzieq/blob/448b6d996a913b698d51bf1cb49671d3d6a7f5e8/suzieq/poller/controller/source/netbox.py#L76, netbox checks if `token` and `url` are present in the config.

This is not the best method to do the validation but it works. After the alpha, we were thinking about using JSONSchema fot validation

suzieq/poller/controller/source/base_source.py
LucaNicosia: I added validation for all fields which are not known for namespaces, devices, auths and sources. If an argument is not declared. an error will be shown

suzieq/poller/controller/source/ansible.py
LucaNicosia: We restored the ansible source to the version which is now used on the 0.15 a part from the 'panos' check. I don't get what you mean with 'allow device autodection'

ddutt: We use ssh to autodetect a device type. If that provides an answer, don't override it with whatever Ansible provides.

suzieq/poller/controller/credential_loader/base_credential_loader.py
ddutt: Aren't name (and type?) mandatory for each plugin

suzieq/poller/controller/source/base_source.py
ddutt: Excellent

suzieq/poller/controller/credential_loader/base_credential_loader.py
LucaNicosia: Think about the validation as 2 steps:
- the first step is to validate the inventory as is. Here we check if the structure of each object config is right. For example checking if the name is specified (type is not mandatory because both sources and auth has a default type)
- the second step is to validate each component. In order to do this, we need to create the objects and each one of them runs a validation before starting. Here we validate that the users doesn't put a value which is not used by object

PR #526, opened 2021-12-27
Bunch of patches to better support IOSXE platform

ddutt:
Based on data from real world by a user.

### DIFF
--- a/suzieq/engines/pandas/interfaces.py
+++ b/suzieq/engines/pandas/interfaces.py
@@ -1,4 +1,3 @@
-from itertools import repeat
 from ipaddress import ip_network
 
 from suzieq.engines.pandas.engineobj import SqPandasEngine
@@ -271,7 +270,7 @@ class InterfacesObj(SqPandasEngine):
                     .fillna({i: [] for i in if_df.index})
 
         if 'vlanList' not in if_df.columns:
-            if_df['vlanList'] = [[] for i in repeat(None, len(if_df))]
+            if_df['vlanList'] = [[] for i in range(len(if_df))]
 
         if lldp_df.empty:
             if status != 'pass':
--- a/suzieq/engines/pandas/interfaces.py
+++ b/suzieq/engines/pandas/interfaces.py
@@ -1,3 +1,4 @@
+from itertools import repeat
 from ipaddress import ip_network
 
 from suzieq.engines.pandas.engineobj import SqPandasEngine
@@ -270,7 +271,7 @@ class InterfacesObj(SqPandasEngine):
                     .fillna({i: [] for i in if_df.index})
 
         if 'vlanList' not in if_df.columns:
-            if_df['vlanList'] = []
+            if_df['vlanList'] = [[] for i in repeat(None, len(if_df))]
 
         if lldp_df.empty:
             if status != 'pass':
--- a/suzieq/poller/worker/coalescer_launcher.py
+++ b/suzieq/poller/worker/coalescer_launcher.py
@@ -7,6 +7,8 @@ import fcntl
 import logging
 import os
 import signal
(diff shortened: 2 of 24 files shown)


suzieq/poller/worker/services/interfaces.py
```diff
@@ -824,21 +836,30 @@ def _clean_iosxr_data(self, processed_data, raw_data):
...
+            entry['type'] = iftype
             # This is specifically for IOSXE/IOS devices where
             # the IPv6 address uses capital letters
             if devtype != 'iosxr':
                 entry['ip6AddressList'] = [x.lower()
                                            for x in entry.get('ip6AddressList',
                                                               [])]
 
+            if entry['ifname'].endswith('.0'):
+                entry['vlan'] = -1
```
LucaNicosia, line 855: Use a constant defined inside utils instead of '-1'?

ddutt: Ack, I agree

suzieq/engines/pandas/interfaces.py
```diff
@@ -270,7 +271,7 @@ def _check_ipaddr(x, fld1, fld2, reason):
                     .fillna({i: [] for i in if_df.index})
 
         if 'vlanList' not in if_df.columns:
-            if_df['vlanList'] = []
+            if_df['vlanList'] = [[] for i in repeat(None, len(if_df))]
```
claudiolor, line 274:
```suggestion
            if_df['vlanList'] = list(repeat([], len(if_df)))
```
If we use itertools something like this should be enough.
Or without using any extension:
```suggestion
            if_df['vlanList'] = [[] for i in range(len(if_df))]
```

suzieq/poller/worker/coalescer_launcher.py
```diff
@@ -135,7 +137,7 @@ async def _start_coalescer(self) -> Process:
         else:
             coalescer_args_str = ''
         coalescer_args_str = f'{self.coalescer_bin} {coalescer_args_str}'
-        coalescer_args = coalescer_args_str.strip().split()
+        coalescer_args = shlex.split(coalescer_args_str.strip())
```
claudiolor, line 140: If the problem is building the list from the string, wouldn't be better to directly create a list? So having something like this:
```python
coalescer_args = [self.coalescer_bin]
if self.config_file:
   coalescer_args += ['-c', self.config_file]
```

suzieq/poller/worker/services/vlan.py
```diff
@@ -1,4 +1,5 @@
 import numpy as np
+from collections import defaultdict
```
claudious96, line 2: This import seems to be unused

suzieq/poller/worker/services/macs.py
```diff
@@ -150,10 +150,18 @@ def _clean_nxos_data(self, processed_data, _):
 
     def _clean_iosxe_data(self, processed_data, _):
 
+        foo = ''
```
claudious96, line 153: Leftover here ?

claudious96 (approved):

suzieq/engines/pandas/interfaces.py
ddutt: Ack

suzieq/poller/worker/coalescer_launcher.py
ddutt: Sure, that'd work too. This seems simpler. The user in this case had a dir with weird characters in it such as (, \, space etc.

suzieq/poller/worker/services/vlan.py
ddutt: Yes, it was removed

PR #529, opened 2021-12-29
Poller debug mode and worker and coalescer log in stdout

claudiolor:
This PR makes the poller debug simpler, introducing the following features:
- When `log-stdout` is true for poller in the configuration file, the log of each worker is shown in the stdout
- When `log-stdout` is true for the coalescer, the stdout of the poller shows the log of the coalescer
- Add `run-once=debug` allowing to generate the inventory without automatically starting the workers, i.e.:
  ```
  $ sq-poller -I ./sq-inventory.yml -w 2 --run-once debug
  2 inventory chunks generated, you can now manually start the workers with the following commands:

  export SQ_CONTROLLER_POLLER_CRED=vtMcqttxk5_2AZW9BNK4XbtV2FTxO_a2D46DrKAs1v0=
  export SQ_INVENTORY_PATH=/tmp/.suzieq/inventory.59752
  python /suzieq/poller/worker/sq_worker.py --worker-id 0

  export SQ_CONTROLLER_POLLER_CRED=vtMcqttxk5_2AZW9BNK4XbtV2FTxO_a2D46DrKAs1v0=
  export SQ_INVENTORY_PATH=/tmp/.suzieq/inventory.59752
  python /suzieq/poller/worker/sq_worker.py --worker-id 1
  ```

### DIFF
--- a/suzieq/poller/controller/controller.py
+++ b/suzieq/poller/controller/controller.py
@@ -56,6 +56,10 @@ class Controller:
         #                    inventory from a source
         self._run_once = args.run_once
 
+        # If the debug mode is active we need to run the controller only once
+        if args.debug:
+            self._run_once = 'debug'
+
         self._input_dir = args.input_dir
         if self._input_dir:
             self._run_once = 'input-dir'
@@ -103,6 +107,7 @@ class Controller:
 
         manager_args = {'config': sq_get_config_file(args.config),
                         'config-dict': config_data,
+                        'debug': args.debug,
                         'input-dir': self._input_dir,
                         'exclude-services': args.exclude_services,
                         'no-coalescer': self._no_coalescer,
--- a/suzieq/poller/controller/controller.py
+++ b/suzieq/poller/controller/controller.py
@@ -108,6 +108,9 @@ class Controller:
                         'no-coalescer': self._no_coalescer,
                         'output-dir': args.output_dir,
                         'outputs': args.outputs,
+                        # We are intentionally passing the run_once argument
+                        # since we do not want to give internal values as
+                        # arguments of the poller workers
                         'run-once': args.run_once,
                         'service-only': args.service_only,
                         'ssh-config-file': args.ssh_config_file,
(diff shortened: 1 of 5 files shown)


suzieq/poller/controller/controller.py
```diff
@@ -108,6 +108,9 @@ def __init__(self, args: argparse.Namespace, config_data: dict) -> None:
                         'no-coalescer': self._no_coalescer,
                         'output-dir': args.output_dir,
                         'outputs': args.outputs,
+                        # We are intentionally passing the run_once argument
+                        # since we do not want to give internal values as
+                        # arguments of the poller workers
```
ddutt, line 113: Claudio, why do you want to use run-once argument. Its not even running once in this case. Why not use --debug?

suzieq/poller/sq_poller.py
```diff
@@ -116,11 +116,14 @@ def controller_main():
...
-              'processing on the data. Both cases store the results in a '
-              'plain output file, one for each service.')
+        choices=['gather', 'process', 'debug'],
+        help=('''The poller do not run forever, three modes are available:
+        (1) gather: store the output as it has been collected,
+        (2) process: performs some processing on the data.
+        Both cases store the results in a plain output file,
+        one for each service, and exit.
+        (3) debug: build the device list and exit without polling
+        the nodes.''')
```
ddutt, line 126: As I said earlier, why not use --debug? This is more of a question, not a request to change the way you're doing it.

suzieq/poller/worker/coalescer_launcher.py
```diff
@@ -64,7 +60,7 @@ async def start_and_monitor_coalescer(self):
         """
         try:
             await self._monitor_coalescer()
-        except CancelledError:
+        except asyncio.CancelledError:
```
ddutt, line 63: I had done the code the way it was deliberately. When we upgrade to 3.8 or whatever version supports exceptions the way you had originally done it, this would automatically work. Unless you object strongly, I suggest we go back to how it was (module import inside a try/except).

ddutt (changes requested):

PR #532, opened 2021-12-30
Fix #429

### LINKED ISSUE #429: Add snapshot mode
**Is your feature request related to a problem? Please describe.**
Today, when the user needs to run suzieq poller just once to gather data, say for network validation, they have a more complex workflow to execute. They have to first run the poller with --run-once=gather option followed by another run of suzieq poller with -i option on the directory the gather was done.

It'd be useful to add a mode called --snapshot that takes the option of running once but writing to parquet directly thereby avoiding the two-step workflow described above. 

It'd be even more useful if the snapshot mode could take a timeout parameter to try executing the command exactly once for each host in the inventory, but add a timeout to give up trying.

**Describe the solution you'd like**
--run-once=gather was a option addded to Suzieq to make it easier for the dev team to support poller parsing issues in the field, and to develop reproducible data for tests. 

**Additional context**

ddutt:
Add a snapshot mode to poller. This has multiple scenarios in which it is useful.

### DIFF
--- a/suzieq/poller/worker/services/service.py
+++ b/suzieq/poller/worker/services/service.py
@@ -62,7 +62,7 @@ class Service(SqPlugin):
         self._poller_schema = value
 
     def __init__(self, name, defn, period, stype, keys, ignore_fields, schema,
-                 queue, run_once="forever"):
+                 queue, run_once=None):
         self.name = name
         self.defn = defn
         self.ignore_fields = ignore_fields or []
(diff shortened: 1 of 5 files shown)


suzieq/poller/worker/services/service.py
```diff
@@ -813,6 +813,11 @@ async def run(self):
             # Post a cmd to fire up the next poll after the specified period
             self.logger.debug(
                 f"Rescheduling service for {self.name} service")
+            if self.run_once == "update":
```
claudiolor, line 816: Just a detail: logging `Rescheduling service` when the mode is `update` is not correct

claudiolor (approved):

suzieq/poller/worker/services/service.py
```diff
@@ -761,7 +761,7 @@ async def run(self):
                 await self.commit_data(result, output[0]["namespace"],
                                        hostname)
             else:
```
claudious96, line 763: Just a small comment:
```suggestion
            elif  self.run_once in ["gather", "process", "update"]:
```

claudious96 (approved):

suzieq/poller/worker/services/service.py
ddutt: Ack, good catch. FIxing it.

suzieq/poller/worker/services/service.py
ddutt: Ack

PR #534, opened 2021-12-31
Fix #533: Support for QFX10K

### LINKED ISSUE #533: Juniper QFX10k polling errors
**Describe the bug**
QFX10K devices are not polled. Logs report error parsing the uptime, and the following modules fail with a bunch of polling errors: 
* device
* evpnVni
* routes
* ospfNbr

**To Reproduce**
1. point a poller at a QFX10K device (e.g. QFX10002-36q) and poll as normal.

**Attach**
[sq-poller.log](https://github.com/netenglabs/suzieq/files/7796496/sq-poller.log)

 **Expected behavior**
Polling should complete and populate the database

**Additional context**
discussed in slack [here](https://networktocode.slack.com/archives/C018DBNH3ST/p1640893892062800)

ddutt:
Junos QFX10K is a platform thats a cross between MX and QFX platforms, and as such deserves its own NOS model. This patch addresses that.

### DIFF
--- a/suzieq/config/devconfig.yml
+++ b/suzieq/config/devconfig.yml
@@ -46,6 +46,9 @@ apply:
   junos-es:
     copy: junos-qfx
 
+  junos-qfx10k:
+    copy: junos-qfx
+
   sonic:
     version: all
     command: show runningconfiguration all
--- a/suzieq/config/arpnd.yml
+++ b/suzieq/config/arpnd.yml
@@ -88,6 +88,9 @@ apply:
   junos-es:
     copy: junos-qfx
 
+  junos-qfx10k:
+    copy: junos-qfx
+
   nxos:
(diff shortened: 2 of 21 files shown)


suzieq/config/devconfig.yml
```diff
@@ -1,5 +1,5 @@
 service: devconfig
-period: 3600
+period: 30
```
claudious96, line 2: Is this a debug leftover? Why change this default?

claudiolor (approved):

suzieq/config/devconfig.yml
ddutt: Ugh. Good catch, will fix before merging. Thanks.

LucaNicosia (approved):

PR #535, opened 2022-01-02
Adding check for ansible_host and ansible_user keys existing

LucaNicosia:
This fix simply checks if the ansible file contains all the correct `ansible_host` and `ansible_user`. Before the fix, if one of them doesn't existed, the code will launch a KeyError

### DIFF
--- a/suzieq/poller/controller/source/ansible.py
+++ b/suzieq/poller/controller/source/ansible.py
@@ -23,8 +23,8 @@ class AnsibleInventory(Source):
         super().__init__(input_data)
 
     def _load(self, input_data):
-        self.ansible_file = input_data.pop('path', None)
-        self.namespace = input_data.pop('namespace', None)
+        self.ansible_file = input_data.get('path', None)
+        self.namespace = input_data.get('namespace', None)
         inventory = self._get_inventory()
         self.set_inventory(inventory)
 
@@ -77,8 +77,20 @@ class AnsibleInventory(Source):
         for host in in_hosts:
             entry = in_hosts[host]
 
+            ansible_host = entry.get('ansible_host')
+            if not ansible_host:
+                logger.warning(f'{self._name} skipping ansible device '
+                               'without name')
+                continue
+
+            ansible_user = entry.get('ansible_user')
+            if not ansible_user:
+                logger.warning(
+                    f'{self._name} skipping ansible device without username')
+                continue
+
             # Get password if any
-            password = ''
+            password = None
             if 'ansible_password' in entry:
                 password = entry['ansible_password']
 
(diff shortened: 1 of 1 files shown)


claudious96 (approved):

suzieq/poller/controller/source/native.py
```diff
@@ -30,6 +31,9 @@ def _validate_config(self, input_data: dict):
         if not isinstance(input_data.get('hosts'), list):
```
claudiolor, line 31: You are here checking if the host list is actually a list, but if the `hosts` is missing, it always return: `Hosts must be a list`. Now I think it is better moving the check at line 34 before, so that if hosts is not defined or it is an empty list it returns `False`, and we might return something like: `The field hosts with the list of nodes to poll must be provided`, then we can check if the `hosts` field is actually a list

LucaNicosia: You right. I fix it right now

suzieq/poller/controller/source/native.py
```diff
@@ -90,6 +91,13 @@ def _get_inventory(self) -> Dict:
                     logger.error(f'Ignoring node {address}')
                     continue
 
+                if keyfile and not Path(keyfile).exists():
+                    logger.warning(
+                        f"Ignored host {address} not existing "
+                        f'keyfile: {keyfile}'
+                    )
```
ddutt, line 101: Same error message fix for missing keyfile as in the ansible case.

suzieq/poller/controller/source/native.py
```diff
@@ -28,11 +28,14 @@ def _validate_config(self, input_data: dict):
...
-        if not isinstance(input_data.get('hosts'), list):
-            raise InventorySourceError(f'{self._name} Hosts must be a list')
-
         if not input_data.get('hosts'):
-            raise InventorySourceError(f'{self._name} Hosts must not be empty')
+            raise InventorySourceError(f"{self._name} The 'hosts' field with "
+                                       "the list of nodes to poll is mandatory"
+                                       )
+
+        if not isinstance(input_data.get('hosts'), list):
```
ddutt, line 36: Same error as what @claudiolor pointed out in the other case? You have to check its not an empty list?

ddutt (changes requested):

suzieq/poller/controller/source/native.py
LucaNicosia: The check is done in the line above. I'm doing `if not input_data.get('hosts'): #error`. This check wiil catch both invalid and empty list.

ddutt (approved):

PR #536, opened 2022-01-03
Fix broken test

claudious96:
This PR fixes the broken interfaces parsing integration test. 
It also defines a more specific policy of when to run github actions:
- on every push in branches that have an open PR
- on every push event on the master branch and on version specific branches (i.e. `0.16.0`)

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -1,5 +1,15 @@
 name: integration-tests
-on: [pull_request]
+on:
+  push:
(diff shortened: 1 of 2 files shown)


claudiolor (approved):

ddutt:
@claudious96 The fix you've made is incorrect. The test is failing because I need to regenerate the ospf-ibgp test data! There was a bug in cumulus interfaces parsing w.r.t. bond interfaces that I have fixed and its in already.

PR #542, opened 2022-01-05
Fix bug in cli show

claudious96:
This PR fixes an unexpected breaking of the CLI that appeared only in interactive mode. It was not caught by integration tests because the tests run each command in non-interactive mode, which does not produce any error.

Example to reproduce:
1. Edit suzieq-cfg file to use `data-directory: ./tests/data/junos/parquet-out/`
2. open the suzieq cli and run: `path show dest='172.16.3.202' src='172.16.1.101' namespace=junos`
3. output: 
```
claudious> path show dest='172.16.3.202' src='172.16.1.101' namespace=junos
Error running command: 'Index' object has no attribute 'remove'
------------------------------------------------------------
Traceback (most recent call last):
  File "/home/claudious/.cache/pypoetry/virtualenvs/suzieq-C5KIDAcn-py3.8/lib/python3.8/site-packages/nubia/internal/cmdbase.py", line 408, in run_interactive
    ret = fn(**args_dict)
  File "/home/claudious/stardust/suzieq/suzieq/cli/sqcmds/command.py", line 182, in show
    return self._gen_output(df)
  File "/home/claudious/stardust/suzieq/suzieq/cli/sqcmds/command.py", line 452, in _gen_output
    cols.remove('timestamp')
AttributeError: 'Index' object has no attribute 'remove'
------------------------------------------------------------

```

### DIFF
--- a/tests/unit/test_cli_command.py
+++ b/tests/unit/test_cli_command.py
@@ -1,0 +1,81 @@
+import pytest
+import pandas as pd
+
+from suzieq.cli.sqcmds.command import SqCommand
+from suzieq.sqobjects.path import PathObj
+
+# valid dataframe extracted from test data
+df = pd.DataFrame({
+    'pathid': {0: 1, 1: 1, 2: 1, 3: 1, 4: 1, 5: 2, 6: 2, 7: 2, 8: 2, 9: 2},
+    'hopCount': {0: 0, 1: 1, 2: 2, 3: 3, 4: 4, 5: 0, 6: 1, 7: 2, 8: 3, 9: 4},
+    'namespace': {0: 'junos', 1: 'junos', 2: 'junos', 3: 'junos', 4: 'junos',
+                  5: 'junos', 6: 'junos', 7: 'junos', 8: 'junos', 9: 'junos'},
+    'hostname': {0: 'server101', 1: 'leaf01', 2: 'spine01', 3: 'leaf02',
+                 4: 'server202', 5: 'server101', 6: 'leaf01', 7: 'spine02',
+                 8: 'leaf02', 9: 'server202'},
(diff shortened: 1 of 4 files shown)


ddutt:
There's one more place in the code that you need to fix. Look at the start of _gen_output under the if 'error'

ddutt:
Do you check the order of the lines? If not, Is it possible to capture the output with output=text and run it via test_sqcmds? (see common-samples/help.yml for an example of this. If you're planning to do this another way with a different set of tests, that works too. Otherwise, I apologize for making you do more on this already excellent work w.r.t. adding a test.

tests/unit/test_cli_command.py
```diff
@@ -0,0 +1,81 @@
...
+                                          'devconfig', ''])
+def test_cli_outputs_with_error(ouput_format, setup_nubia):
+    '''Test output generation with errors'''
+
+    sqcmd = SqCommand(format=ouput_format, sqobj=PathObj)
+    assert sqcmd._gen_output(df) == 1
+
+
+@pytest.mark.cli
+@pytest.mark.parametrize('ouput_format', ['json', 'csv', 'markdown',
```
ddutt, line 75: Typo in parameter name, "ouput" instead of "output"

ddutt (changes requested):

claudious96:
> Do you check the order of the lines? If not, Is it possible to capture the output with output=text and run it via test_sqcmds? (see common-samples/help.yml for an example of this. If you're planning to do this another way with a different set of tests, that works too. Otherwise, I apologize for making you do more on this already excellent work w.r.t. adding a test.

@ddutt actually this test only checks whether the function `_gen_output()` returns correctly. The test passes as long as no exception is thrown and the return code is 0 in case of no errors and 1 in case of errors.

I'll add some related cases to sqcmds in the common-samples folder since this is not specific for a single command/nos.

PR #545, opened 2022-01-06
Removing deprecated ssl validation

LucaNicosia:
While testing I found that the parameter `verify_ssl` is deprecated. You need to use `ssl` instead.

### DIFF
--- a/suzieq/poller/controller/source/netbox.py
+++ b/suzieq/poller/controller/source/netbox.py
@@ -115,10 +115,16 @@ class Netbox(Source, InventoryAsyncPlugin):
             headers ([dict]): headers to initialize the session
         """
         if not self._session:
-            self._session = aiohttp.ClientSession(
-                headers=headers,
-                connector=aiohttp.TCPConnector(verify_ssl=self._ssl_verify)
-            )
+            if self._ssl_verify:
+                self._session = aiohttp.ClientSession(
+                    headers=headers,
+                    connector=aiohttp.TCPConnector(ssl=None)
+                )
+            else:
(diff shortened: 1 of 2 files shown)


suzieq/poller/controller/source/netbox.py
```diff
@@ -115,10 +115,16 @@ def _init_session(self, headers: dict):
...
+            if self._ssl_verify:
+                self._session = aiohttp.ClientSession(
+                    headers=headers,
+                    connector=aiohttp.TCPConnector(ssl=None)
+                )
+            else:
+                self._session = aiohttp.ClientSession(
+                    headers=headers,
+                    connector=aiohttp.TCPConnector(ssl=False)
+                )
```
ddutt, line 127: Why not simply set ssl to self._ssl_verify?

ddutt (changes requested):

suzieq/poller/controller/source/netbox.py
LucaNicosia: I set the ssl=False instead of _ssl_verify because it was easier to read. The only way the code will enter in this code is with self._ssl_verify = False

claudiolor, line 129:
```suggestion
            ssl_option = None if self._ssl_verify else False
            self._session = aiohttp.ClientSession(
                headers=headers,
                connector=aiohttp.TCPConnector(ssl=ssl_option)
            )
```
Agree with Dinesh, I think is better having something like this

claudious96 (approved):

ddutt (approved):

PR #551, opened 2022-01-11
Pylint fixes last

ddutt:
Hopefully the last set of pylint fixes. We should be fully pylint compliant at this point.


.github/workflows/linting.yml
```diff
@@ -15,6 +15,26 @@ jobs:
           python-version: '3.8'
           architecture: 'x64'
 
+      - name: Install Poetry
```
claudious96, line 18: The pre-commit check will keep failing even if the code is compliant because the pre-commit framework doesn't allow to run a poetry install before running a hook (i.e. pylint). Therefore we see the import errors and the action failed.
The pylint check should be moved to another job unrelated to pre-commit so that we can actually install poetry before lint the code.

tests/integration/test_coalescer.py
```diff
@@ -195,7 +197,7 @@ async def _run(cmd):
         stdout=asyncio.subprocess.PIPE,
         stderr=asyncio.subprocess.PIPE)
 
-    stdout, stderr = await proc.communicate()
+    _, _2 = await proc.communicate()
```
LucaNicosia, line 200: I think here you can also write `await proc.communicate()` and the linter doesn't complain

claudiolor (approved):

tests/integration/test_coalescer.py
ddutt: True enough, will fix.

.github/workflows/linting.yml
ddutt: Yes, I agree.

PR #553, opened 2022-01-11
Improve coalescer monitoring

claudiolor:
Currently when there is another instance of the coalescer running, the coalescer launcher task terminates without logging anything. The aim of this PR is improving the way in which the coalescer is monitored:
- Removed the file lock acquire: this check is not needed, provided we monitor the status of the coalescer process. Indeed, whenever there is another instance of the coalescer running, the launched coalescer terminates with a EBUSY error state. Moreover, not acquiring the lock prevents the poller from waiting some seconds before checking the status of the coalescer.
- Set a maximum number of attempts for coalescer startup: when the coalescer terminates for a reason different than "another coalescer instance running", we increase an attempt counter and retry, until the counter reaches the maximum value

### DIFF
--- a/suzieq/poller/worker/coalescer_launcher.py
+++ b/suzieq/poller/worker/coalescer_launcher.py
@@ -3,17 +3,15 @@ This module contains all the logic needed to start and monitor the coalescer
 """
 import asyncio
 import errno
-import fcntl
 import logging
-import os
-import signal
 from pathlib import Path
 import shlex
 from asyncio.subprocess import Process
 from typing import Dict
 
 from suzieq.poller.controller.utils.proc_utils import monitor_process
-from suzieq.shared.utils import ensure_single_instance, get_sq_install_dir
+from suzieq.shared.exceptions import PollingError
+from suzieq.shared.utils import get_sq_install_dir
 
 logger = logging.getLogger(__name__)
 
@@ -26,7 +24,8 @@ class CoalescerLauncher:
     def __init__(self,
                  config_file: str,
                  cfg: Dict,
-                 coalescer_bin: str = None) -> None:
+                 coalescer_bin: str = None,
+                 max_attempts: int = 3) -> None:
         """Initialize an instance of the CoalescerLauncher
 
         Args:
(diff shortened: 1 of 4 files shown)


suzieq/poller/worker/coalescer_launcher.py
```diff
@@ -78,48 +87,45 @@ async def _monitor_coalescer(self):
         """This function calls _start_coalescer() and check if the process
         dies
         """
-        fd = 0
-        # Check to see file lock is possible
-        while not fd:
+
+        starting_attemps = 0
```
ddutt, line 91: Typo in name: attemps instead of attempts. Typos can lead to strange, unexpected errors. Please fix it.

suzieq/poller/worker/coalescer_launcher.py
```diff
@@ -78,48 +87,45 @@ async def _monitor_coalescer(self):
...
-                                        False)
-            if fd > 0:
-                # unlock and try to start process
-                try:
-                    fcntl.flock(fd, fcntl.F_UNLCK)
-                    os.close(fd)
-                except OSError:
-                    pass
-                continue
+                raise PollingError('Unable to start the coalescer')
```
ddutt, line 106: Lets not print this unless there's no coalescer and we can't start it. If there's a coalescer already running, printing this error may scare users.

tests/unit/poller/test_coalescer_launcher.py
```diff
@@ -97,7 +91,48 @@ async def test_coalescer_start(coalescer_cfg):
...
+            patch.multiple(coalescer_launcher_module,
+                           monitor_process=monitor_process_fn):
+            await asyncio.wait_for(cl.start_and_monitor_coalescer(), 5)
+        # Check if multiple attempts have been performed
+        attempts_done = start_coalescer_fn.call_count
+        assert attempts_done == cl.max_attempts, \
+            f'Expected {cl.max_attempts} attempts, {attempts_done} done'
+    except asyncio.TimeoutError:
+        pytest.fail(
+            'The coalescer launcher task expected to file but it does not')
```
ddutt, line 124: typo. "..expected to fail", not "..expected to file"

ddutt (changes requested):

suzieq/poller/worker/coalescer_launcher.py
ddutt: Ack, thx for clarifying.

ddutt (approved):

PR #554, opened 2022-01-11
Linting action

claudious96:
This PR refactor the liniting pipeline, creating a job dedicated to pylint and flake8 and a pre-commit action dedicated to other kind of checks (such as newline at end of files).


.github/workflows/pre-commit.yml
```diff
@@ -0,0 +1,22 @@
+name: pre-commit
+
+on: [pull_request]
```
ddutt, line 3: Do we need to add what branches this applies to like the previous action (lines 6-8)?

ddutt (changes requested):

.github/workflows/pre-commit.yml
claudious96: I'm not sure, if we are going to keep the pre-commit action read-only then yes. Otherwise, if it will have the ability to modify the commit I think it is better not to have it on push events since it might be unexpected.

claudious96: On second thoughts, since now it is read-only it's okay to have it on push events too. We will think about it when and if it will be given the ability to modify the commit.

PR #557, opened 2022-01-11
sq-poller: gracefully terminate with -i mode

claudiolor:
Currently when the poller is launched with the `-i` flag, it raises an exception when the worker terminates. This PR solves this problem.

### DIFF
--- a/suzieq/poller/controller/controller.py
+++ b/suzieq/poller/controller/controller.py
@@ -48,26 +48,27 @@ class Controller:
         self._config.update(config_data.get('poller', {}))
 
         # Set controller configuration
-        # run_once: ['gather', 'process', 'update'] tells the controller if
-        #           the poller should query the devices and exit
+        # single_run_mode: ['gather', 'process', 'update', 'debug',
+        #                   'input-dir']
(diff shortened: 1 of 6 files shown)


ddutt (approved):

ddutt:
The one issue with your fix is that we cannot use the run-once=gather output to then be used for run-once=debug. Isn't that right?

LucaNicosia (approved):

claudious96 (approved):

ddutt (changes requested):
I see what you did based on our Slack discussion. But do we need both run-once and single-run-mode the way its coded?

claudiolor:
> I see what you did based on our Slack discussion. But do we need both run-once and single-run-mode the way its coded?

I think so, because they are different things, `single-run-mode` is an internal variable, which we use inside the controller. We keep `run_once` because we need to pass it as is to the worker. Furthermore, they might assume different values, for example when we launch the controller with the `--debug` flag  and we set `--run-once=update`, `single-run-mode` will be equal to `debug` while `run-once=update` and we need it to show the right command to execute.

PR #558, opened 2022-01-12
Fix #523: Make help string consistent, unambiguous

### LINKED ISSUE #523: Improved help verbiage when using the hostname option
**Describe the bug**
When using the ```hostname=``` qualifier in commands like ```device show``` incorrectly, the help text states:
*hostname must be a space separated list of strings*
which some might interpret as hostname=["dev1", "dev2", "dev3"] 
when in fact what is needed is hostname="dev", "dev2", "dev3

**To Reproduce**
1. Invoke the ```suzieq-cli
2. Execute the following command and observe the output
```
claudia> device show hostname=127.0.0.1
hostname must be a space separated list of strings
This command does not support positional arguments
claudia>
```
3. In addition, the "pop up" help states  *Space separated list of hostnames to qualify*

**Attach**

 **Expected behavior**
Consistent help text in both the resulting error "tip" and the "pop up help" (is it space or comma or are both supported?)

Something along the lines of:
hostname should be a quoted string (hostname="dev1").  separate multiple hostnames with a comma (hostname="dev1", "dev2"
..or something along those lines.  The use of the word list should be avoided unless an actual python list is required (or include examples in the help text)

**Screenshots**
image

**Additional context**
Technically this is not a bug but more of a suggestion!

ddutt:
Fix for 523 along with test fixes.

### DIFF
--- a/suzieq/cli/sqcmds/NetworkCmd.py
+++ b/suzieq/cli/sqcmds/NetworkCmd.py
@@ -36,10 +36,10 @@ class NetworkCmd(SqCommand):
         )
 
     @command("show", help="Show device information")
-    @argument("model", description="models to filter with")
-    @argument("os", description='NOS to filter with')
-    @argument('vendor', description='vendor to filter with')
-    @argument('version', description='NOS version to filter with')
+    @argument("model", description="model(s), space separated")
+    @argument("os", description='NOS(es), space separated')
+    @argument('vendor', description='Vendor(s), space separated')
+    @argument('version', description='NOS version(s), space separated')
     # pylint: disable=arguments-differ
     def show(self, os: str = "", vendor: str = "", model: str = "",
              version: str = "") -> int:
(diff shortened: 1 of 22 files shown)


suzieq/cli/sqcmds/NetworkCmd.py
```diff
@@ -59,7 +59,7 @@ def show(self, os: str = "", vendor: str = "", model: str = "",
         return self._gen_output(df)
 
     @command("find", help="find where an IP/MAC address is attached")
-    @argument("address", type=str, description="IP/MAC address to find")
+    @argument("address", type=str, description="IP/MAC addresses, in quotes, space separated")
```
LucaNicosia, line 62: Please just split this line in 2 because the linter is complaining
```suggestion
    @argument("address", type=str,
              description="IP/MAC addresses, in quotes, space separated")
```

claudious96 (approved):

suzieq/cli/sqcmds/NetworkCmd.py
ddutt: Fixed. I saw the linter fail.

claudiolor (approved):

PR #569, opened 2022-01-17
Remove warning on path gui page

LucaNicosia:
This PR removes warning from the gui path page and fixes a bug on the same page: performing a valid trace, going to another page and coming back to the path page results into a page fail


ddutt (changes requested):
I don't see the problem that you describe happening.

PR #570, opened 2022-01-18
Move describe table in xplore page

LucaNicosia:
This PR moves the help of each table from the sidebar to the bottom of the table in the xplore page

### DIFF
--- a/suzieq/gui/stlit/guiutils.py
+++ b/suzieq/gui/stlit/guiutils.py
@@ -1,13 +1,13 @@
-from importlib.util import find_spec
+import base64
 import os
 from enum import Enum
-import base64
+from importlib.util import find_spec
 
-import streamlit as st
 import pandas as pd
-from streamlit.server.server import Server
+import streamlit as st
+from IPython.display import Markdown
 from streamlit.report_thread import get_report_ctx
-
+from streamlit.server.server import Server
 from suzieq.sqobjects import get_sqobject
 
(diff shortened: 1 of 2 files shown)


suzieq/gui/stlit/xplore.py
```diff
@@ -114,10 +115,10 @@ def _create_sidebar(self) -> None:
             state.namespace = namespace
 
         if state.table:
-            tables_obj = get_sqobject('tables')(start_time=state.start_time,
+            state.tables_obj = get_sqobject('tables')(start_time=state.start_time,
```
ddutt: You can use the describe function of the table itself. So you can do `get_sqobject(state.table)().describe()`. I'm going to deprecate `table describe command=` at some point.

suzieq/gui/stlit/guiutils.py
```diff
@@ -1,13 +1,13 @@
-from importlib.util import find_spec
+import base64
 import os
 from enum import Enum
-import base64
+from importlib.util import find_spec
```
ddutt, line 4: I'm confused. Did you add these? We shouldn't be needing them, IIRC.

ddutt (changes requested):

suzieq/gui/stlit/guiutils.py
LucaNicosia: I didn't added them. It is only import reordering.
base64 is used by the image and find_spec is used by the function `get_image_dir`

PR #580, opened 2022-01-26
Poller worker minor fixes

claudiolor:
This PR contains some minor fixes for the poller worker:
- Rename the Poller class to Worker
- Allow to pass additional arguments to the Inventory init function
- Currently when the poller is not able to connect to all the nodes of the inventory the following error is reported:
  ```Terminating because no nodes or services found```
  This error seems to be misleading. This PR shows different errors for the nodes and the services cases.
- Add a clearer message when the host key is unknown

### DIFF
--- a/suzieq/poller/worker/nodes/node.py
+++ b/suzieq/poller/worker/nodes/node.py
@@ -365,8 +365,15 @@ class Node:
                 if self.sigend:
                     await self._terminate()
                     return
-                self.logger.error("ERROR: Unable to connect to "
-                                  f"{self.address}:{self.port}, {e}")
+                if isinstance(e, asyncssh.HostKeyNotVerifiable):
+                    self.logger.error(
+                        f'Unable to connect to {self.address}: {self.port}, '
+                        'host key is unknown. If you do not need to verify '
+                        'the host identity, add "ignore-known-hosts: True" in '
+                        'the device section of the inventory')
+                else:
+                    self.logger.error('Unable to connect to '
+                                      f'{self.address}:{self.port}, {e}')
                 self.current_exception = e
                 await self._close_connection()
                 if rel_lock:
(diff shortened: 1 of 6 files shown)


claudious96 (approved):

LucaNicosia (approved):

suzieq/poller/worker/poller.py
```diff
@@ -153,8 +178,11 @@ def _init_inventory(self, userargs, cfg):
             'ssh_config_file': userargs.ssh_config_file,
         }
 
+        if addnl_args:
+            inventory_args.update(addnl_args)
+
         # Retrieve the specific inventory source to use
-        inv_types = Inventory.get_plugins()
+        inv_types = self._get_inventory_classes()
```
ddutt, line 197: Whats the motivation to add a new function and change the name?

suzieq/poller/worker/nodes/node.py
```diff
@@ -365,8 +365,15 @@ async def _init_ssh(self, init_boot_time=True, rel_lock=True) -> None:
                 if self.sigend:
                     await self._terminate()
                     return
-                self.logger.error("ERROR: Unable to connect to "
-                                  f"{self.address}:{self.port}, {e}")
+                if isinstance(e, asyncssh.HostKeyNotVerifiable):
+                    self.logger.error(
+                        f'Unable to connect to {self.address}: {self.port}, '
+                        'host key is unknown. If you do not need to verify '
+                        'the host identity, add "ignore-known-host: True" in '
```
ddutt, line 372: The keyword is "ignore-known-hosts", not "ignore-known-host".

ddutt (changes requested):
One MUST FIX before commit and the others are just comments or q.

suzieq/poller/worker/poller.py
ddutt: Yes, it does, thank you

PR #581, opened 2022-01-26
Netbox env token + inventory validation

LucaNicosia:
This PR:
- adds the possibility to specify the netbox API token via an environment variable
- improve inventory validation

The validation of the inventory is more advanced than a syntactical once:
- validates also the hosts in the native format
- validate ansible input file
- validate netbox ssl-verify parameter is not True for http connections
- validate credential files
- validate env variables
- no additional parameters are allowed

### DIFF
--- a/docs/inventory.md
+++ b/docs/inventory.md
@@ -71,7 +71,7 @@ namespaces:
   auth: credentials-from-file-0
 ```
 
-## <a name='sensitive-data'></a>Sensible data
+## <a name='sensitive-data'></a>Sensitive data
 A sensitive data is an information that the user doesn't want to store in plain-text inside the inventory.
 For this reason, SuzieQ inventory now supports three different options to store these kind of informations
 
--- a/docs/inventory.md
+++ b/docs/inventory.md
@@ -71,6 +71,15 @@ namespaces:
   auth: credentials-from-file-0
 ```
 
+## <a name='sensitive-data'></a>Sensible data
+A sensitive data is an information that the user doesn't want to store in plain-text inside the inventory.
+For this reason, SuzieQ inventory now supports three different options to store these kind of informations
+
+- `plain:<password-in-plaintext>` or `<password-in-plaintext>`: the sensitive information is stored as is in the inventory
+- `env:<ENV_VARIABLE>`: the sensitive information is stored in an environment variable
+- `ask`: the user can write the sensitive information on the stdin
+
+Currently this method is used to specify passwords, passphrases and tokens.
 ## <a name='inventory-sources'></a>Sources
 
 The device sources currently supported are:
@@ -132,7 +141,11 @@ Now you can set the path of the ansible inventory in the source:
 
 ### <a name='source-netbox'></a>Netbox
 
-[Netbox](https://netbox.readthedocs.io/en/stable/) is often used to store devices type, management address and other useful information to be used in network automation. Suzieq can pull device data from Netbox selecting them by tag (currently only one per each source). To do so a token to access the netbox API is required as well as the netbox instance url.
+[Netbox](https://netbox.readthedocs.io/en/stable/) is often used to store devices type, management address and other useful information to be used in network automation. Suzieq can pull device data from Netbox selecting them by tag (currently only one per each source).
+To do so a token to access the netbox API is required as well as the netbox instance url.
+
+The token is considered a [sensitive data](#sensitive-data), so it can be specified via an environment variable using the format `env:ENV_TOKEN`.
+
 Since Netbox is a _dynamic source_, the data are periodically pulled, the period can be set to any desired number in seconds (default is 3600).
 
 !!!Info
@@ -172,7 +185,7 @@ sources:
 auths:
 - name: auth-st
   username: user
-  passowrd: my-password
+  password: my-password
 
 namespaces:
 - name: netbox-sitename # devices namespaces equal to their site names
@@ -232,12 +245,6 @@ The simplest method is defining either username and password/private key.
   password: plain:pass
 ```
 
-where `password` can be specified in plaintext, as an environment variable or to be asked to the user:
-
-- `plain:<password-in-plaintext>`
-- `env:<ENV_VARIABLE>`
-- `ask`
-
 In case a private key is used to authenticate:
 ```yaml
 - name: suzieq-user
@@ -245,7 +252,10 @@ In case a private key is used to authenticate:
   key-passphrase: ask
 ```
 
-Where `key-passphrase` is the passphrase of the private key. As the `password` field, it can be set as plaintext, env variable or to be asked to the user.
+Where `key-passphrase` is the passphrase of the private key.
+
+Both `passoword` and `key-passphrase` are considered [sensitive data](#sensitive-data).
+For this reason they can be set as plaintext, env variable or asked to the user via stdin.
 
 ### <a name='cred-file'></a>Credential file
 
--- a/suzieq/poller/controller/credential_loader/base_credential_loader.py
+++ b/suzieq/poller/controller/credential_loader/base_credential_loader.py
@@ -3,12 +3,36 @@ devices credentials
 """
 import logging
 from abc import abstractmethod
-from typing import Dict, List, Type
+from typing import Dict, List
 
-from suzieq.poller.controller.base_controller_plugin import ControllerPlugin
+from suzieq.poller.controller.base_controller_plugin import \
+    InventoryPluginModel, ControllerPlugin
 from suzieq.shared.exceptions import InventorySourceError
 
 logger = logging.getLogger(__name__)
+
+
+def check_credentials(inventory: Dict):
+    """Checks if all the credentials are set
+    """
+
+    # check if all devices has credentials
+    no_cred_nodes = [
+        f"{d.get('namespace')}.{d.get('address')}"
+        for d in inventory.values()
+        if not d.get('username', None) or
+        not (d.get('password') or d.get('ssh_keyfile'))
+    ]
+    if no_cred_nodes:
+        raise InventorySourceError(
+            'No credentials to log into the following nodes: '
+            f'{no_cred_nodes}'
+        )
+
+
+class CredentialLoaderModel(InventoryPluginModel):
+    """Model for credential loader validation
+    """
 
 
 class CredentialLoader(ControllerPlugin):
@@ -16,8 +40,8 @@ class CredentialLoader(ControllerPlugin):
     sources
     """
 
-    def __init__(self, init_data: Dict) -> None:
-        super().__init__()
+    def __init__(self, init_data: Dict, validate: bool = True) -> None:
+        super().__init__(init_data, validate)
 
         self._cred_format = [
             'username',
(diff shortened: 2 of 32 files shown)


suzieq/poller/controller/utils/inventory_utils.py
```diff
@@ -161,3 +180,17 @@ def get_inventory_config(conf_type: str, inventory: dict) -> dict:
...
+
+    Args:
+        name (str): environment variable to search
+        default (str, optional): value returned if the variable doesn't exists
+        Defaults to ''.
+
+    Returns:
+        str: environment variable or default
+    """
+    return getenv(name, default)
```
claudious96, line 196: I don't see any advantages in wrapping `getenv` in `get_env_var`. Is it just to have a `default = ''`?

LucaNicosia: It is more a styling modification. I don't like to see the import for `os.getenv` everywhere

suzieq/poller/controller/utils/inventory_utils.py
```diff
@@ -62,7 +63,9 @@ def validate_raw_inventory(inventory: dict):
     if not inventory:
         raise InventorySourceError('The inventory is empty')
 
-    for f in ['sources', 'namespaces']:
+    mandatory_fields = ['sources', 'namespaces']
+
+    for f in mandatory_fields:
         if f not in inventory:
```
claudiolor, line 69:
```suggestion
    if any(m not in inventory for m in mandatory_fields):
```

suzieq/poller/controller/utils/inventory_utils.py
```diff
@@ -71,7 +74,23 @@ def validate_raw_inventory(inventory: dict):
         'sources': [],
         'devices': [],
         'auths': [],
+        'namespaces': []
     }
+
+    invalid_fields = [
+        x for x in inventory
+        if x not in list(main_fields.keys()) + ['namespaces']]
```
claudiolor, line 82: `namespace` is already in `main_fields`

suzieq/poller/controller/credential_loader/static.py
```diff
@@ -42,7 +42,7 @@ def init(self, init_data: dict):
         if init_data.get('password'):
             password = init_data['password']
             if password.startswith('env:'):
```
claudiolor, line 44: What about having a function in utils doing this work? So that whenever we need to get some confidential data we can call it without repeating the same code. i.e.
```
password = get_pwd(init_data['password'])
```

suzieq/poller/controller/source/netbox.py
```diff
@@ -68,7 +69,14 @@ def _load(self, input_data: dict):
         self._namespace = input_data.get('namespace')
         self._period = input_data.get('period', 3600)
         self._run_once = input_data.get('run_once', False)
-        self._token = input_data.get('token')
+        token = input_data.get('token')
+        if token.startswith('env:'):
```
claudiolor, line 73: As said in the previous comment we can have this logic in a function

suzieq/poller/controller/utils/inventory_utils.py
```diff
@@ -161,3 +180,17 @@ def get_inventory_config(conf_type: str, inventory: dict) -> dict:
                 configs[name] = conf_obj
 
     return configs
+
+
+def get_env_var(name: str, default: str = '') -> str:
```
claudiolor, line 185: Better to have a function for parsing the password here instead of one getting the environment variable

suzieq/poller/controller/utils/inventory_utils.py
```diff
@@ -161,3 +180,35 @@ def get_inventory_config(conf_type: str, inventory: dict) -> dict:
...
+
+    Raises:
+        InventorySourceError: environment variable not found
+
+    Returns:
+        str: sensible data
+    """
+    sens_data = input_method
+    if input_method.startswith('env:'):
+        sens_data = getenv(input_method.split('env:')[1], '')
```
claudiolor, line 205: Shouldn't we use `strip()` to remove leading and trailing whitespaces? If for example we write `env: VAR` instead of `env:VAR`

suzieq/poller/controller/utils/inventory_utils.py
```diff
@@ -161,3 +180,35 @@ def get_inventory_config(conf_type: str, inventory: dict) -> dict:
...
+        if not sens_data:
+            raise InventorySourceError(
+                f'No environment variable called '
+                f"'{input_method.split('env:')[1]}'")
+    elif input_method.startswith('plain:'):
+        sens_data = input_method.split("plain:")[1]
```
claudiolor, line 211: Same problem of spaces as before

docs/inventory.md
```diff
@@ -132,7 +141,12 @@ Now you can set the path of the ansible inventory in the source:
 
 ### <a name='source-netbox'></a>Netbox
 
-[Netbox](https://netbox.readthedocs.io/en/stable/) is often used to store devices type, management address and other useful information to be used in network automation. Suzieq can pull device data from Netbox selecting them by tag (currently only one per each source). To do so a token to access the netbox API is required as well as the netbox instance url.
+[Netbox](https://netbox.readthedocs.io/en/stable/) is often used to store devices type, management address and other useful information to be used in network automation. Suzieq can pull device data from Netbox selecting them by tag (currently only one per each source).
+To do so a token to access the netbox API is required as well as the netbox instance url.
+
+The token is considered a [sensible data](#sensible-data), so it can be specified via an environment variable.
+The input via the stdin is not supported.
```
claudiolor, line 148: Looking at the code it seems you actually support it, don't you?

docs/inventory.md
```diff
@@ -135,7 +144,8 @@ Now you can set the path of the ansible inventory in the source:
 [Netbox](https://netbox.readthedocs.io/en/stable/) is often used to store devices type, management address and other useful information to be used in network automation. Suzieq can pull device data from Netbox selecting them by tag (currently only one per each source).
 To do so a token to access the netbox API is required as well as the netbox instance url.
 
-The token can be directly written in the inventory or it can be stored in an environment variable and indicated using the format `env:<ENV_TOKEN>`.
+The token is considered a [sensible data](#sensible-data), so it can be specified via an environment variable.
+The input via the stdin is not supported.
```
ddutt: I liked the statement about "env: <ENV_TOKEN>". Why did you take it out?

ddutt:
Can we add an option to merely syntax check the inventory file? So sq-poller --syntax-check -I <inventory file>. If you have a different option string you prefer, let me know.

ddutt (changes requested):

docs/inventory.md
LucaNicosia: I thought it was a repetition since I wrote that it is a sensible data, but I added it back. Writing the same thing twice sometimes is better

docs/inventory.md
```diff
@@ -71,6 +71,15 @@ namespaces:
   auth: credentials-from-file-0
 ```
 
+## <a name='sensitive-data'></a>Sensible data
+A sensitive data is an information that the user doesn't want to store in plain-text inside the inventory.
+For this reason, SuzieQ inventory now supports three different options to store these kind of informations
+
+- `plain:<password-in-plaintext>` or `<passowrd-in-plaintext>`: the sensitive information is stored as is in the inventory
```
claudious96, line 78: typo
```suggestion
- `plain:<password-in-plaintext>` or `<password-in-plaintext>`: the sensitive information is stored as is in the inventory
```

docs/inventory.md
```diff
@@ -132,7 +141,11 @@ Now you can set the path of the ansible inventory in the source:
 
 ### <a name='source-netbox'></a>Netbox
 
-[Netbox](https://netbox.readthedocs.io/en/stable/) is often used to store devices type, management address and other useful information to be used in network automation. Suzieq can pull device data from Netbox selecting them by tag (currently only one per each source). To do so a token to access the netbox API is required as well as the netbox instance url.
+[Netbox](https://netbox.readthedocs.io/en/stable/) is often used to store devices type, management address and other useful information to be used in network automation. Suzieq can pull device data from Netbox selecting them by tag (currently only one per each source).
```
claudious96, line 144:
```suggestion
[Netbox](https://netbox.readthedocs.io/en/stable/) is often used to store devices type, management address and other useful information to be used in network automation. SuzieQ can pull device data from Netbox selecting them by tag (currently only one per each source).
```

docs/inventory.md
```diff
@@ -132,7 +141,11 @@ Now you can set the path of the ansible inventory in the source:
...
+To do so a token to access the netbox API is required as well as the netbox instance url.
+
+The token is considered a [sensitive data](#sensitive-data), so it can be specified via an environment variable using the format `env:ENV_TOKEN`.
```
claudious96, line 147: Just to be consistent
```suggestion
The token is considered a [sensitive data](#sensitive-data), so it can be specified via an environment variable using the format `env:<ENV_TOKEN>`.
```

docs/inventory.md
```diff
@@ -232,20 +245,17 @@ The simplest method is defining either username and password/private key.
...
 ```yaml
 - name: suzieq-user
   keyfile: path/to/private/key
   key-passphrase: ask
 ```
 
-Where `key-passphrase` is the passphrase of the private key. As the `password` field, it can be set as plaintext, env variable or to be asked to the user.
+Where `key-passphrase` is the passphrase of the private key.
+
+Both `passoword` and `key-passphrase` are considered [sensitive data](#sensitive-data).
```
claudious96, line 257:
```suggestion
Both `password` and `key-passphrase` are considered [sensitive data](#sensitive-data).
```

suzieq/poller/controller/credential_loader/static.py
```diff
@@ -34,53 +49,39 @@ def init(self, init_data: dict):
...
+        if not self._validate:
+            # fix pydantic alias
+            init_data['ssh_passphrase'] = init_data.pop('ssh-passphrase', None)
+        super().init(init_data)
+
+        if self._data.password == 'ask':
+            try:
+                self._data.password = get_sensitive_data(
+                    self._data.password,
+                    f'{self.name} Password to login to device: ')
```
claudious96, line 61: I think here we should print also some device info to identify which one we are asking the password for (hostname, address, maybe even the namespace).

docs/inventory.md
```diff
@@ -71,6 +71,15 @@ namespaces:
   auth: credentials-from-file-0
 ```
 
+## <a name='sensitive-data'></a>Sensible data
```
claudious96, line 74:
```suggestion
## <a name='sensitive-data'></a>Sensitive data
```

claudious96 (changes requested):

claudiolor (approved):

suzieq/poller/controller/credential_loader/base_credential_loader.py
```diff
@@ -26,23 +50,39 @@ def __init__(self, init_data: Dict) -> None:
             'passphrase'
         ]
 
-        # load auth parameters
+        self._data: CredentialLoaderModel = None
+        self.init(init_data)
 
-        self._name = init_data.get('name')
+    @property
+    def name(self) -> str:
```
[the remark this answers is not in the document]
LucaNicosia: For now it is not possible due to the missing chunker and manager validation. After we add that we can move it

suzieq/poller/controller/credential_loader/static.py
LucaNicosia: It is not true. This is the static loader, so you are specifying a single password to access all the devices

PR #584, opened 2022-01-30
Route & Path improvements

ddutt:
This handles the fixes to stop resolving routes during routes show (since its only ever used for path), and therefore handling it in path.


suzieq/engines/pandas/path.py
```diff
@@ -400,6 +428,51 @@ def _get_underlay_nexthop(self, hostname: str, vtep_list: list,
...
+                                hostname=[row.hostname], address=[nhop],
+                                vrf=[row.vrf])
+
+                if not rrdf.empty:
+                    if rrdf.oifs.tolist() == []:
+                        # Recurse once more
+                        max_recurse -= 1
+                        if max_recurse:
+                            rrdf = self._handle_recursive_route(rrdf,
+                                                                max_recurse)
```
LucaNicosia, line 459: Suppose that we have `row.nexthopIps` with 3 elements. Everyone of them has a route table such as `rrdf` is not empty. Suppose also that `max_recurse = 2`.
Doing `max_recurse -= 1` will never trigger the recursion for the 3rd element of `row.nexthopIps` and I don't think it is what you expected. I suggest to remove the `max_recurse -= 1` and write `rrdf = self._handle_recursive_route(rrdf, max_recurse-1)`

suzieq/engines/pandas/routes.py
```diff
@@ -85,6 +57,12 @@ def get(self, **kwargs):
             addnl_fields.append('prefixlen')
             drop_cols.append('prefixlen')
 
+        if 'oifs' in columns:
+            for x in ['namespace', 'hostname', 'vrf', 'prefix', 'nexthopIps']:
```
claudiolor, line 61: Why we need to retrieve these additional fields when oifs is in columns?

claudious96 (approved):
+1 to Luca's suggestion. Overall looks good to me.

suzieq/engines/pandas/path.py
```diff
@@ -422,23 +495,21 @@ def _get_nexthops(self, device: str, vrf: str, dest: str, is_l2: bool,
         if not rslt.empty and (len(rslt.nexthopIps.iloc[0]) != 0 and
                                rslt.nexthopIps.iloc[0][0] != '') and (
                                    rslt.protocol.iloc[0] != 'hmm'):
+            if not rslt.empty and (len(rslt.nexthopIps.iloc[0]) != 0 and
```
claudiolor, line 498: We are checking two times the following statement: `not rslt.empty and (len(rslt.nexthopIps.iloc[0]) != 0 and rslt.nexthopIps.iloc[0][0] != '')`. Why not merging the two if statements?

suzieq/engines/pandas/path.py
```diff
@@ -422,23 +495,21 @@ def _get_nexthops(self, device: str, vrf: str, dest: str, is_l2: bool,
...
+                                   rslt.nexthopIps.iloc[0][0] != '') and (
+                    not rslt.oifs.iloc[0].tolist()):
```
claudiolor, line 500: We might use `.size` instead of converting to list every time

suzieq/engines/pandas/path.py
```diff
@@ -400,6 +428,51 @@ def _get_underlay_nexthop(self, hostname: str, vtep_list: list,
...
+            oifs = []
+            nhops = []
+            for nhop in row.nexthopIps:
+                rrdf = self._get_table_sqobj('routes') \
+                           .lpm(namespace=[row.namespace],
+                                hostname=[row.hostname], address=[nhop],
+                                vrf=[row.vrf])
+
+                if not rrdf.empty:
+                    if rrdf.oifs.tolist() == []:
```
claudiolor, line 454: As the previous comment we might avoid to always convert to list. We might use `.size` or, since this is a Pandas series even  `.empty`

suzieq/engines/pandas/routes.py
```diff
@@ -85,6 +57,12 @@ def get(self, **kwargs):
             addnl_fields.append('prefixlen')
```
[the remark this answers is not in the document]
ddutt: Because I add prefixlen only if necessary, not otherwise whereas the others are added anyway

suzieq/engines/pandas/routes.py
```diff
@@ -134,9 +112,6 @@ def get(self, **kwargs):
             srs_max = np.amax(srs, 1)
             df.insert(len(df.columns)-1, 'numNexthops', srs_max)
 
-        if not df.empty and 'oifs' in df.columns:
```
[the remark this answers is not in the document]
ddutt: Not necessarily. There are people who want to see it where they're expecting to see it. Not all routes are without oifs.

suzieq/engines/pandas/path.py
```diff
@@ -119,8 +126,16 @@ def _init_dfs(self, ns, source, dest):
             # TODO: No host with this src addr. Is addr a local ARP entry?
```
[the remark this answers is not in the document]
ddutt: Yes, Fixed

Issue #458, opened 2021-10-30
Suzieq shows bgp sessions which has been removed from the device


rasanentimo:
**Describe the bug**
Suzieq shows bgp sessions which has been removed from the device

**To Reproduce**
1. create bgp configuration
2. poll the data with suzieq
3. delete bgp configration
4. the data is still shown
5. delete older parquet files
6. the data is still shown

I'm not sure if this requires some other steps like restarting the poller at some phase.

**Attach**
n/a

 **Expected behavior**
Expecting the bgp session not be visible in suzieq after it has been removed from device

Expecting the bgp session not be visible in suzieq after the older parquet files have been deleted. 

**Screenshots**
n/a

**Additional context**
This bgp session is not present on the device but it's shown on suzieq

```
         peer    state                  estdTime                        timestamp
0  Ethernet39  NotEstd 1970-01-01 00:00:00+00:00 2021-10-09 10:29:41.690000+00:00
```
the parquet files doesn't exist for that date either

```
# ls -lah
total 40M
drwxr-xr-x 2 root root  40M Oct 29 20:02 .
drwxr-xr-x 5 root root  83M Sep 23 11:00 ..
-rw-r--r-- 1 root root  92K Oct 10 00:01 sqc-h-1633820400-1633824000.parquet
-rw-r--r-- 1 root root 102K Oct 10 01:01 sqc-h-1633824000-1633827600.parquet
-rw-r--r-- 1 root root 102K Oct 10 02:01 sqc-h-1633827600-1633831200.parquet
-rw-r--r-- 1 root root 102K Oct 10 03:01 sqc-h-1633831200-1633834800.parquet
-rw-r--r-- 1 root root 102K Oct 10 04:01 sqc-h-1633834800-1633838400.parquet
-rw-r--r-- 1 root root  92K Oct 10 05:01 sqc-h-1633838400-1633842000.parquet
-rw-r--r-- 1 root root  92K Oct 10 06:01 sqc-h-1633842000-1633845600.parquet
-rw-r--r-- 1 root root  92K Oct 10 07:01 sqc-h-1633845600-1633849200.parquet
```

ddutt:
[...]

Issue #388, opened 2021-07-15
Duplicate Records on Poller Restart


rickdonato:
**Describe the bug**
When the poller is restarted the tables show duplicate entries.

**To Reproduce**
1. restart poller. 
2. look at the table with a `view=all`

**Attach**
N/A

 **Expected behavior**
The duplicates are removed on startup so only differences are seen within the table.

**Screenshots**
N/A

**Additional context**
N/A

ddutt:
Fixed in 0.17.1

PR #598, opened 2022-02-04
Remove old records

claudiolor:
This PR addresses the issues #458 and #388. 
Whenever a service receives the result of a poll, it won’t write everything every time, but it checks all the differences with the previous poll. If the result is the same then nothing is written. This approach allows to minimize the amount of data written by the poller, but has a great shortcoming: when we do not have any record explicitly invalidating  the previous ones with active=False, we assume that it is valid. This results in: 
- Records staying forever because the poller was offline when they have been removed
- Duplicate records when the poller restart, since it doesn't have the previous state

Required steps:

- [x] Fix get_diff() function so that it is independent from the order of the fields
- [x] Restore the state of the previous polls when the poller starts up


claudious96 (approved):

suzieq/poller/worker/services/service.py
```diff
@@ -572,9 +595,28 @@ def clean_data_common(self, processed_data, raw_data):
...
-        prev_res = self.previous_results.get(key, [])
+        prev_res = self.previous_results.get(key, None)
+        # No data previously polled with this service from the the
+        # current namespace and hostname. Check if the datastore contains some
+        # information about
+        filter_df = f"hostname.str.match('{hostname}') and " \
+                    f"namespace.str.match('{namespace}')"
+        if prev_res is None:
+            df = self._db_access.read(
+                self.schema_table._table,
```
LucaNicosia, line 609: Since the function `_get_db_access` may return None. Shouldn't we add a check here for self._db_access?

claudiolor: We use db_access only when we actually use a DB. When we _get_db_access() returns None we are not using a dB, so we'll never enter this function

Issue #573, opened 2022-01-24
Support for "first-seen" / "last-seen" timestamp


LumirH:
I believe SuzieQ needs to store not only timestamp when new record was created, but also timestamp when the record was last updated. 

I will demonstrate this on "network find" API, but the problem is generic. Lets assume this method/API call:

root> network find address="10.1.1.1" view=all format=markdown
|    | namespace   | hostname               | vrf     | ipAddress   |   vlan | macaddr           | ifname                  | bondMembers   | type   | l2miss   | timestamp                        |
|---:|:------------|:-----------------------|:--------|:------------|-------:|:------------------|:------------------------|:--------------|:-------|:---------|:---------------------------------|
|  0 | site | router1 | default | 10.1.1.1 |      0 | cc:7f:76:54:5f:33 | GigabitEthernet0/1/3 |               | routed | False    | 2022-01-13 15:40:03.817000+00:00 |
|  1 | site     | router2 | default | 10.1.1.1 |      0 | cc:7f:76:54:34:21 | GigabitEthernet0/1/3 |               | routed | False    | 2022-01-18 16:44:16.415000+00:00 |

From this output, I can say following facts about the network:
- On 13th January, IP address 10.1.1.1 was "seen" on router1 port GigabitEthernet0/1/3
- On 18th January, IP address 10.1.1.1 was "seen" to router2 port GigabitEthernet0/1/3

I would be interested in additional information:
- Is IP address 10.1.1.1 still available on router2 ? When it was last seen on this port ? How old is the information that I see in this output ? 
- Was IP address 10.1.1.1 connected to router1 for example on 15th January ? (device with this IP address could have been disconnected from the network for example on 15th January and reconnected on 18th January to router2)

**Proposed solution:**
Store two timestamps. First will be updated when record has been first created (first-seen), second timestamp would be updated every time when you re-discover the same information.

ddutt:
This is fixed without the suggestion. Here is a sample output seen in 0.17.1:
```
ddutt> network find start-time='4 Apr 2022, 11 pm' end-time='5 Apr 2022, 5:30 am' address='172.16.2.201'
  active namespace hostname       vrf     ipAddress vlan            macaddr  ifname bondMembers     type  l2miss                        timestamp
0      +      demo   leaf03  evpn-vrf  172.16.2.201   20  fe:95:fd:1a:1e:2e  bond01        swp3  bridged   False 2022-04-04 23:18:30.769000-07:00
[...]

PR #599, opened 2022-02-04
Add lastPoll column

claudious96:
This PR partially addresses #573 adding to all the tables the new augmented column `lastPoll` that shows the timestamp of the last successful polling event for that record. In case there is no polling data yet, 0 is shown (this is possible in case of run once and no data present in the sqPoller table). The column is not shown by default, only when specified with `columns='*'` or `columns='lastPoll'`.

Tasks:

- [x] add augmented column to table schemas
- [x] write logic to get the lastPoll timestamp from the sqPoller table
- [x] manual testing adding and removing nodes
- [ ] update test data

### DIFF
--- a/suzieq/engines/pandas/device.py
+++ b/suzieq/engines/pandas/device.py
@@ -55,25 +55,6 @@ class DeviceObj(SqPandasEngine):
         if view == 'latest' and 'status' in df.columns:
             df['status'] = np.where(df.active, df['status'], 'dead')
 
-        if columns == ['*'] or 'lastSeenTimestamp' in columns:
-            # consider only polling events with passed status
-            lastSeen_df = self._get_table_sqobj('sqPoller').get(
-                namespace=kwargs.get('namespace', []),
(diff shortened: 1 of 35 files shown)


claudiolor (approved):

suzieq/engines/pandas/device.py
```diff
@@ -55,6 +55,25 @@ def get(self, **kwargs):
         if view == 'latest' and 'status' in df.columns:
             df['status'] = np.where(df.active, df['status'], 'dead')
 
+        if columns == ['*'] or 'lastSeenTimestamp' in columns:
+            # consider only polling events with passed status
+            lastSeen_df = self._get_table_sqobj('sqPoller').get(
+                namespace=kwargs.get('namespace', []),
+                hostname=kwargs.get('hostname', []),
+                service='device',
+                columns='namespace hostname status'.split(),
```
LucaNicosia, line 64: You can change it to `['namespace', 'hostname', 'status']`.

Issue #576, opened 2022-01-25
NX-OS SFP inventory data in incorrect


anubisg1:
**Describe the bug**
inventory show displays "type: XCVR" for every SFP installed instead of the real SFP type

 **Expected behavior**
the SFP type should be displayed

**additional information**

inventory.yml file polls this command:

```
    - command: show interface transceiver | json native
      normalize: 'TABLE_interface/ROW_interface/*/[
      "interface: name?|",
      "sfp: _sfp?|",
      "name: model?|",
      "type: type?|",
      "rev: version?|",
      "partnum: partNum?|",
      "serialnum: serial?|",
      "cisco_product_id: productId?|",
      ]'
```

this is the switch output 

```
swich_01 # show interface transceiver | json native {
		"TABLE_interface": {
			"ROW_interface": [{
						"interface": "Ethernet1/1",
						"sfp": "present",
						"type": "10Gbase-SR",
						"name": "CISCO-AVAGO",
						"partnum": "SFBR-709SMZ-CS2",
						"rev": "G4.1",
						"serialnum": "AVD2203DE7L",
						"nom_bitrate": 10300,
						"len_50": 82,
						"len_625": 26,
						"len_50_OM3": 300,
						"ciscoid": "3",
						"ciscoid_1": 4,
						"cisco_part_number": "10-3105-01",
						"cisco_product_id": "SFP-10G-SR-S",
						"cisco_vendor_id": "V01"
					},
```

notice that "type" is   "10Gbase-SR", but suzieq only shows xcvr 

![image](https://user-images.githubusercontent.com/4518931/150971360-e627e36f-cbf4-4991-80ae-88dec029eb2e.png)

ddutt:
This info is shown via partType. Fixed in 0.17.1

Issue #568, opened 2022-01-15
v0.16.0 docker image REST server fails to allow connections


jeremyschulman:
**Describe the bug**

When using the docker container latest v0.16.0 and attempting to follow the instructions found here:
https://suzieq.readthedocs.io/en/latest/rest-server/

I was not able to connect with the server.  After troubleshooting the problem with Dinesh, the remediation was
to update within the container the file `/root/.suzieq/suzieq-cfg.yml` and changing the rest.address value from 127.0.0.1 to 0.0.0.0.

The documentation also indicates when starting the REST server to invoke the sq-rest-server.py file, but as of v0.16.0 the file name is simply sq-rest-server.

Thank you.

PR #602, opened 2022-02-11
Misc patches

ddutt:
There's a bunch of patches in this, the primary one being the support for portmode and vlanList. There's also a couple of patches such as adding the ability to connect to the rest server from outside the container (#568), fixing inventory info on NXOS (#576)

I had these patches all done in one place to get one of the users going. Apologies for mixing them.


ddutt:
I'll fix the merge issues on the tests after you've reviewed the fixes. I did a git rebase which caused a bunch of patches to get added here which are not really part of the patch, but actually other commits that got added again because of the rebase. I'm not sure why this happened and how to avoid it the next time, except to not rebase the next time. I rebased to incorporate the fixes made to put all tests under a single dir

suzieq/engines/pandas/interfaces.py
```diff
@@ -35,16 +55,45 @@ def get(self, **kwargs):
...
+                if x in columns or '*' in columns:
+                    continue
+                drop_cols.append(x)
+            df = self._add_portmode(df)
+
+        if vlan or "vlanList" in columns or '*' in columns:
+            if 'portmode' not in columns and '*' not in columns:
+                for x in ['ipAddressList', 'ip6AddressList']:
+                    if x in columns or '*' in columns:
+                        continue
```
LucaNicosia, line 69: At line 68, the `'*' in columns` is useless since you can enter there only if `'*' not in columns`

suzieq/engines/pandas/interfaces.py
```diff
@@ -425,3 +470,145 @@ def _check_ipaddr(x, fld1, fld2, reason):
...
+                # Heuristics now
+                if '\ninterfaces {\n' in row.config or \
+                   'paloaltonetworks' in row.config:
+                    syntax = 'junos'
+                else:
+                    syntax = 'ios'
+            try:
+                conf = CiscoConfParse(row.config.split('\n'), syntax=syntax)
+            except Exception:  # pylint: disable=broad-except
+                break
```
LucaNicosia, line 527: This for is highly dependent on the order due to the `break` at line 527.
Suppose you have 3 elements in conf_df: the first one is invalid (raise the CiscoConfParse exception) and the other 2 are valid. The result is that none of the is processes.
If the same 3 elements are in a different order [valid, invalid, valid], now only the first one is processed.
Using `continue` instead of break should fix this

Dockerfile
```diff
@@ -11,6 +11,7 @@ WORKDIR /suzieq
 COPY ./dist/suzieq-$version-py3-none-any.whl  /tmp/
 RUN pip install /tmp//suzieq-$version-py3-none-any.whl
 COPY suzieq/config/etc/suzieq-cfg.yml /root/.suzieq/suzieq-cfg.yml
+RUN sed -i 's/127.0.0.1/0.0.0.0/' /root/.suzieq/suzieq-cfg.yml
```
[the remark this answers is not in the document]
ddutt: Because with 127.0.0.1 only enabled inside a docker container, nobody outside the container can connect to it

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "suzieq"
-version = "0.16.0"
+version = "0.16.1"
```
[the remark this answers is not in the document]
ddutt: Yep, will fix

suzieq/engines/pandas/interfaces.py
ddutt: Good catch, will fix

suzieq/engines/pandas/interfaces.py
ddutt: Another excellent catch, yes

claudious96 (approved):
The pre-commit check is failing, you should just add a newline at the end of the file `tests/integration/sqcmds/junos-input/devconfig.output`.

suzieq/engines/pandas/interfaces.py
```diff
@@ -425,3 +470,145 @@ def _check_ipaddr(x, fld1, fld2, reason):
...
+            .get(namespace=df.namespace.unique().tolist(),
+                 hostname=df.hostname.unique().tolist())
+
+        devdf = self._get_table_sqobj('device') \
+            .get(namespace=df.namespace.unique().tolist(),
+                 hostname=df.hostname.unique().tolist(),
+                 columns=['namespace', 'hostname', 'os', 'status', 'vendor'])
+
+        if conf_df.empty:
+            df['portmode'] = np.where(df.ipAddressList.str.len() == 0,
```
claudiolor, line 494: I think we might avoid to repeat this piece of code by replacing lines [493:500] with:
```
if not conf_df.empty:
```
wrapping from line 502 to 542. And replacing 544 with:
```
if conf_df.empty or pm_df.empty:
```

claudiolor (approved):
In the commits of this PR two different docstring formats have been used. I think we should fix it

suzieq/engines/pandas/interfaces.py
ddutt: I hate deeply nested code. I did a different thing. Since we're iterating over the rows of conf_df. If its empty, pm_list will be empty and therefore, I only need the code on line 544, ```if pm_df.empty```. That addresses your issue as well. Do you agree @claudiolor ?

suzieq/shared/schema.py
```diff
@@ -76,10 +76,36 @@ def key_fields_for_table(self, table: str) -> List[str]:
...
-    def augmented_fields_for_table(self, table: str) -> List[str]:
-        '''Returns list of augmented fields in table'''
-        return [x['name'] for x in self._schema.get(table, [])
-                if 'depends' in x]
+    def augmented_fields_for_table(self, table: str,
+                                   fields: List[str],
+                                   recurse: int) -> List[str]:
+        '''Return all augmented fields in given list of fields (all if empty)
+        If recurse is non-zero, continue to recurse till all fields are
+        resolved
```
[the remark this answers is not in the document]
ddutt: Ack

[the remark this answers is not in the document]
ddutt: Just leaving it consistent with the rest of the code in that file. Otherwise, that routine will look out of place.

suzieq/engines/pandas/interfaces.py
claudiolor: Sure, makes sense

PR #604, opened 2022-02-19
4B ASN support

ddutt:
This adds support for 4B ASN parsing in case the NOS reports the ASN in asndot format, and for displaying the ASN in asndot format.

### DIFF
--- a/suzieq/engines/pandas/bgp.py
+++ b/suzieq/engines/pandas/bgp.py
@@ -28,6 +28,10 @@ class BgpObj(SqPandasEngine):
         sch = self.schema
         fields = sch.get_display_fields(columns)
 
+        if columns == ['*']:
+            fields.remove('sqvers')
+            fields.remove('origPeer')
+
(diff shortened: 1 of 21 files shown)


LucaNicosia (approved):

claudious96 (approved):

suzieq/engines/pandas/bgp.py
```diff
@@ -67,9 +78,9 @@ def get(self, **kwargs):
 
         if query_str:
             return mdf.query(query_str).drop(columns=drop_cols,
-                                             errors='ignore')
+                                             errors='ignore')[fields]
         else:
-            return mdf.drop(columns=drop_cols, errors='ignore')
+            return mdf.drop(columns=drop_cols, errors='ignore')[fields]
```
claudious96, line 83: A test is failing here because "origPeer" is in `fields` but not it is dropped before the column selection.
```suggestion
            return mdf[fields]
```
Same in the `if` statement 2 lines above.

claudious96 (changes requested):

PR #605, opened 2022-02-20
Device fix with filters

ddutt:
Device show with filters has been broken for a while, and without the appropriate tests, we haven't caught it. This PR fixes the code and adds additional tests to help catch those errors the next time.

### DIFF
--- a/suzieq/cli/sqcmds/DeviceCmd.py
+++ b/suzieq/cli/sqcmds/DeviceCmd.py
@@ -1,4 +1,4 @@
-import re
+import shlex
 from nubia import command
 from suzieq.cli.nubia_patch import argument
 
@@ -7,8 +7,7 @@ from suzieq.sqobjects.device import DeviceObj
 
 
 @command("device", help="Act on device data")
-@argument("status", description="filter by polling status",
-          choices=["dead", "alive",  "neverpoll"])
+@argument("status", description="filter by polling status",)
 @argument("os", description="NOS(s), space separated")
 @argument("version", description="NOS version(s), space separated")
 @argument("vendor", description="Vendor(s), space separated")
@@ -45,13 +44,21 @@ class DeviceCmd(SqCommand):
             query_str=query_str,
             sqobj=DeviceObj,
         )
+
+        # Nubia eats up the first and last quote. Detect and add back
+        model = model.strip()
+        if model.count("'") % 2 != 0:
+            if not model.startswith("'"):
+                model = f"'{model}"
+            elif not model.endswith("'"):
+                model = f"{model}'"
         if model:
-            model = re.split(r"\s+(?=[^']*(?:'))", model)
+            model = shlex.split(model)
 
         self.lvars = {
             'os': os.split(),
             'version': version.split(),
-            'status': status,
+            'status': status.split(),
             'model': model,
             'vendor': vendor.split()
         }
--- a/suzieq/engines/pandas/device.py
+++ b/suzieq/engines/pandas/device.py
@@ -25,8 +25,6 @@ class DeviceObj(SqPandasEngine):
         user_query = kwargs.pop('query_str', '')
         status = kwargs.pop('status', '')
         os_version = kwargs.pop('version', '')
-        vendor = kwargs.get('vendor', '')
-        model = kwargs.get('model', '')
         os = kwargs.get('os', '')
 
         drop_cols = []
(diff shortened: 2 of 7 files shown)


suzieq/cli/sqcmds/DeviceCmd.py
```diff
@@ -51,7 +50,7 @@ def __init__(
         self.lvars = {
             'os': os.split(),
             'version': version.split(),
-            'status': status,
+            'status': status.split(),
             'model': model,
```
claudiolor, line 54: We are not accepting multiple values in the `model` field

suzieq/engines/pandas/device.py
```diff
@@ -110,25 +108,20 @@ def get(self, **kwargs):
         # The poller merge kills the filtering we did earlier, so redo:
```
claudiolor, line 108: In line 106 we are returning the dataframe without stripping off the columns to drop, so we show the empty dataframe with the internal columns. Moreover, should we move this check before merging the device df with the poller one? (so before line 56)

ddutt: Nice try, but no dice :) The model split is a little more complicated because model names can have spaces inside them. So the split happens a little above, if you read it more carefully :-p

suzieq/cli/sqcmds/DeviceCmd.py
ddutt: Nice try, but no dice :) The model split is a little more complicated because model names can have spaces inside them. So the split happens a little above, if you read it more carefully :-p

suzieq/engines/pandas/device.py
ddutt: I fixed it to return the fields only requested. However, I left this logic where it is because its possible that we end up with an empty dataframe which might happen if there's no poller info. Should we return the device info even without the poller info? Not sure I agree. I also took the time to clean up some old cruft in deciding what fields to return. Did a git push -f

suzieq/cli/sqcmds/DeviceCmd.py
ddutt: But I did find that my regex didn't work correctly. I used shlex split to handle this case. See the code and additional tests that I've added

LucaNicosia (approved):

suzieq/engines/pandas/device.py
```diff
@@ -105,40 +107,30 @@ def get(self, **kwargs):
...
         df = self._handle_user_query_str(df, user_query)
 
         # if poller has failed completely, Can mess up the order of columns
-        cols = self.iobj.schema.get_display_fields(columns)
-        if columns == ['default'] and 'timestamp' not in cols:
-            cols.append('timestamp')
-        if 'sqvers' in cols:
-            cols.remove('sqvers')
-        return df.drop(columns=drop_cols, errors='ignore')[cols]
+        return df.drop(columns=drop_cols, errors='ignore')[fields]
```
claudious96, line 133: Why drop columns if then columns are picked with `[fields]`?
```suggestion
        return df[fields]
```

claudious96 (commented):
As a personal note, do we really need to handle the "!" for device status? There are only 3 possible values and achieving status="!alive" is pretty simple without using the negation: status="dead neverpoll".

suzieq/cli/sqcmds/DeviceCmd.py
claudiolor: I think we still have a problem if we search for `My device model` this is the array we get `['My', 'device', 'model']`, which I think is fine, since coherent with the rest. So let's add quotes, so we search for `'My device model'`, since Nubia strips the quotes the result is equal to the previous one. Let's now try to search for multiple models `'My device model' 'model-x'`, this is the string we get `My device model' 'model-x` and this is the array we get `['My', 'device', 'model model-x']`, which is completely wrong. Since trailing and leading quotes are stripped, distinguishing between space as separator or as space of the model string is not simple at all. Some possible solution I could come up with are:
1. Always add trailing and leading quotes, but at that point we are forced to quotes when defining multiple models in the filter
2. Using a different separator which is not the space, but this won't be coherent with the rest
3. Users can search for a single model at a time

ddutt: So, he way to search for multiple models is: model=" 'my device model' vm". We'll have to document this somewhere. I agree that this is ridiculous, but we can't do anything until we fix nubia. I don't like 2, and we could live with 3.

PR #610, opened 2022-02-24
ospf: Various fixes including much improved assert

ddutt:
This includes a much improved assert that addresses a bunch of issues in the existing model. It also includes a bug fix for summarize.

Test updates to follow soon

### DIFF
--- a/suzieq/engines/pandas/ospf.py
+++ b/suzieq/engines/pandas/ospf.py
@@ -1,3 +1,4 @@
+from typing import List
 from ipaddress import IPv4Network
 
 import numpy as np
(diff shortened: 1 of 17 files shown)


suzieq/engines/pandas/ospf.py
```diff
@@ -280,41 +290,42 @@ def aver(self, **kwargs):
...
+            .query('~adjState.isin(["full", "passive"]) and '
+                   'ifState == "down"') \
+            .reset_index(drop=True)
+        ifdown_df['assertReason'] += \
+            ifdown_df.ifState.apply(lambda x: ['ifdown'])
+
+        # Validate loopback interfaces are of the right network type
+        passed_df['assertReason'] += passed_df.apply(
+            lambda x: ['network type !loopback']
+            if (any(ele in x.ifname for ele in ['lo', 'oopbac']) and
```
LucaNicosia, line 310: is it right the array `['lo', 'oopbac']` or it is a typo on the second element?

suzieq/engines/pandas/ospf.py
```diff
@@ -280,41 +290,42 @@ def aver(self, **kwargs):
...
 
         # In case of Junos, the OSPF interface name is a subif of
         # the interface name. So, create a new column with the
         # orignal interface name.
-        ospf_df['lldpIfname'] = ospf_df['ifname'].str.split('.').str[0]
-        int_df = ospf_df[use_cols].merge(
+        failed_df['lldpIfname'] = failed_df['ifname'].str.split('.').str[0]
+        use_cols = failed_df.columns
+
+        int_df = failed_df[use_cols].merge(
```
LucaNicosia, line 328: If use_cols == failed_df.columns, I don't think you need to do `failed_df[use_cols]` right?

suzieq/engines/pandas/ospf.py
ddutt: Yes, some NOS have lo as their loopback name, others use Loopback, and some other loopback

suzieq/engines/pandas/ospf.py
ddutt: True enough

suzieq/engines/pandas/ospf.py
```diff
@@ -323,106 +334,170 @@ def aver(self, **kwargs):
...
+            dflist: List[pd.DataFrame]
+            result: str: user specification of what result to return
+
+        Returns:
+            The assert result dataframe
+        """
+        ospf_df = pd.concat(dflist)
+        ospf_df['result'] = ospf_df.assertReason.apply(
+            lambda x: 'fail' if x else 'pass')
+        # Weed out the loopback, SVI interfaces as they have no LLDP peers
```
claudious96, line 495: I think this is a leftover comment after a refactoring, right?

PR #611, opened 2022-02-24
Fix reconnection for panos

claudious96:
Fix for properly handling the reconnection in case of a panos node being offline and then back online. There is also some clean up of unused code.

The node went back in the `init` state while it was not necessary. That also meant never moving back to the `good` state since the token was already set.

### DIFF
--- a/suzieq/poller/worker/nodes/node.py
+++ b/suzieq/poller/worker/nodes/node.py
@@ -1443,6 +1443,7 @@ class PanosNode(Node):
     async def _init(self, **kwargs):
         super()._init(**kwargs)
         self.devtype = "panos"
+        self._session = None
 
     async def init_node(self):
         try:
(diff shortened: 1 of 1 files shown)


suzieq/poller/worker/nodes/node.py
```diff
@@ -1470,90 +1471,21 @@ async def init_node(self):
                         "data": output}]
                     await self._parse_boottime_hostname(res, None)
             self._session = aiohttp.ClientSession(
-                conn_timeout=self.connect_timeout,
-                connector=aiohttp.TCPConnector(ssl=False),
-            )
+                    conn_timeout=self.connect_timeout,
+                    connector=aiohttp.TCPConnector(ssl=False),
```
ddutt, line 1475: Is SSL always False or is this the way you specify to ignore certificate verification?

ddutt (approved):

suzieq/poller/worker/nodes/node.py
claudious96: It is to ignore certificate validation, [reference](https://docs.aiohttp.org/en/stable/client_reference.html#:~:text=SSL%20validation%20mode).

PR #613, opened 2022-02-24
BGP assert fixes

LucaNicosia:
- [x] bgp status to `!dynamic` wasn't working. I fixed it
- [x] bgp default message for bgp sessions with a state different from `Established`
- [ ] I need to know which assertions to trigger when the user calls `bgp assert state=dynamic`.

### DIFF
--- a/suzieq/engines/pandas/bgp.py
+++ b/suzieq/engines/pandas/bgp.py
@@ -216,7 +216,8 @@ class BgpObj(SqPandasEngine):
 
         kwargs.pop("columns", None)  # Loose whatever's passed
         result = kwargs.pop("result", 'all')
-        state = kwargs.pop('state', '!dynamic')
+        # by default, state paramter is set to ''
+        state = kwargs.pop('state', '') or '!dynamic'
 
         df = self.get(columns=assert_cols, state=state, **kwargs)
         if 'error' in df:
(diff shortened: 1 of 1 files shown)


suzieq/engines/pandas/bgp.py
```diff
@@ -272,6 +273,10 @@ def _check_if_state(row, if_df):
                 else [], axis=1)
 
             failed_df['result'] = 'fail'
+            failed_df['assertReason'] += failed_df .apply(
+                lambda x: ['Not established']
+                if len(x.assertReason) == 0 else [],
+                axis=1)
```
ddutt, line 279: Luca, you're doing this to avoid the assertReason being blank? Doesn't this show up in all failures then, as an additional reason to anything else we might find? If so, I'm not sure this is useful. The state field we display should be sufficient to indicate that this failed, and hence the result is fail. Maybe we should set it to 'unknown cause' or something like that.

ddutt (changes requested):

suzieq/engines/pandas/bgp.py
LucaNicosia: Yes. This message is going to be displayed only when the assertion is failing with a blank reason. You right, I'll change it to 'Unknow reason'

ddutt:
Very old PR. Closing it.

PR #614, opened 2022-02-27
Topology fixes

ddutt:
This patch incorporates a bunch of topology fixes based on issues uncovered during a debug session with Andrea.


claudiolor (approved):

claudiolor:
We need to add the additional arguments even in the rest server

LucaNicosia (approved):

claudious96 (approved):

Issue #607, opened 2022-02-21
every table has a different number of "devices" under the summarize table


anubisg1:
every table has a different number of "devices" under the summarize table.

Total number of expected devices is 40, what i see is the following (just some examples below) .
Notice that all tables should be working for all the 40 devices, as those features/information is polled for all

![image](https://user-images.githubusercontent.com/4518931/154917642-6d2533ff-6f41-4bfd-9e0f-1e5af61d4ef1.png)
![image](https://user-images.githubusercontent.com/4518931/154917673-48f8d126-7a39-4205-af36-8dc9de58cd4b.png)
![image](https://user-images.githubusercontent.com/4518931/154917713-4fb56dbb-a944-49a5-bfb3-7e6a0050baa5.png)
![image](https://user-images.githubusercontent.com/4518931/154917818-292df59e-4905-4a0d-9a70-21e11d79405d.png)
![image](https://user-images.githubusercontent.com/4518931/154917880-8181f4a7-2012-4d76-b1bd-35c8a19677d7.png)
![image](https://user-images.githubusercontent.com/4518931/154917532-6c82e6fa-a5c1-4ee3-9286-9d0bf8b1ecbb.png)

anubisg1:
as additional information in logs i see

```
2022-02-16 21:43:05,713 - suzieq.poller.worker.nodes.node - ERROR - show vrf detail | json native output for AAAA.BBBB.com failed due to SSH connection closed
```

for several commands on different devices 

sqpoller show is empty

```
root> sqpoller show
Empty DataFrame
Columns: [namespace, hostname, service, status, gatherTime, totalTime, svcQsize, wrQsize, nodeQsize, rxBytes, pollExcdPeriodCount, active, timestamp]
Index: []
root>
```

no errors are recorded in the cli or the ui , nor it's clear how many devices have connection failures or if suzieq is re-establishing the connection

PR #615, opened 2022-02-27
Add hysteresis for marking record as inactive

claudious96:
This change avoids having records swinging back and forth between active and inactive and causing a wrong device count in summarize (#607).
This might happen in case of:
- slow nodes
- timeouts on busy TACACS server authorizing the cmd
- nodes unreachable just for a polling event

As default, the number of unsuccessful polling event before marking the record as inactive is 2 but it is open to be discussed and to be made configurable.  

### DIFF
--- a/suzieq/poller/worker/services/service.py
+++ b/suzieq/poller/worker/services/service.py
@@ -25,7 +25,7 @@ from suzieq.version import SUZIEQ_VERSION
 # How long b4 declaring node dead
 HOLD_TIME_IN_MSECS = 60000
 # How many unsuccessful polls before marking records with active=false
-HYSTERESIS_FAILURE_CNT = 2
+HYSTERESIS_FAILURE_CNT = 3
 
 
 @dataclass
--- a/suzieq/poller/worker/services/service.py
+++ b/suzieq/poller/worker/services/service.py
@@ -22,7 +22,10 @@ from suzieq.shared.sq_plugin import SqPlugin
 from suzieq.shared.utils import known_devtypes
 from suzieq.version import SUZIEQ_VERSION
 
-HOLD_TIME_IN_MSECS = 60000  # How long b4 declaring node dead
+# How long b4 declaring node dead
+HOLD_TIME_IN_MSECS = 60000
+# How many unsuccessful polls before marking records with active=false
+HYSTERESIS_FAILURE_CNT = 2
 
 
 @dataclass
@@ -88,6 +91,7 @@ class Service(SqPlugin):
         self._poller_schema = {}
         self.node_boot_times = defaultdict(int)
         self._failed_node_set = set()
+        self._consecutive_failures = {}
 
         self.poller_schema = property(
             self.get_poller_schema, self.set_poller_schema)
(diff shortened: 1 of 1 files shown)


LucaNicosia (approved):

suzieq/poller/worker/services/service.py
```diff
@@ -748,6 +758,32 @@ async def run(self):
...
+
+                # we should always commit during the first run of the service
+                # for each node regardless their status.
+                if not self.is_first_run(key):
+                    # If it's not the first run and some command did fail,
+                    # increment the count to be used in hysteresis.
+                    # The info will be committed in a future run.
+                    if any((not Service.is_status_ok(x) for x in ostatus)):
+                        self._consecutive_failures[key] = \
+                            self._consecutive_failures.get(key, 0) + 1
```
claudiolor, line 775: Nit: we might use a `deafultdict()` with default value to 0 for `self._consecutive_failures`, this allow us to reduce this line to:
```
self._consecutive_failures[key] += 1
```
since when you refer to a not existing key, it automatically creates it assigning the default value (0 in this case).
Even line 780 would benefit from this modification.

claudiolor (approved):

suzieq/poller/worker/services/service.py
```diff
@@ -22,7 +22,10 @@
 from suzieq.shared.utils import known_devtypes
 from suzieq.version import SUZIEQ_VERSION
 
-HOLD_TIME_IN_MSECS = 60000  # How long b4 declaring node dead
+# How long b4 declaring node dead
+HOLD_TIME_IN_MSECS = 60000
+# How many unsuccessful polls before marking records with active=false
+HYSTERESIS_FAILURE_CNT = 2
```
ddutt, line 28: Can you make this three please? Why 3? I think just about every protocol I know has a dead time thats 3 times the keep alive time :)

suzieq/poller/worker/services/service.py
```diff
@@ -748,6 +758,32 @@ async def run(self):
...
+                    else:
+                        self._consecutive_failures[key] = 0
+
+                if 0 < self._consecutive_failures.get(key, 0) < \
+                        HYSTERESIS_FAILURE_CNT:
+                    self.logger.info(
+                        f"{self.name} {key} node failure hysteresis,"
+                        "skipping data commit")
+                    should_commit = False
+
```
ddutt, line 781: Are we checking if this is the case due to specific failures such as timeout and ssh connection failed as opposed to any failure? We should get into hysterisis only on connectivity failures.

ddutt (changes requested):

suzieq/poller/worker/services/service.py
claudious96: I thought one would be too few but no specific reason for choosing 2. I'll make it 3.

suzieq/poller/worker/services/service.py
claudious96: It's not easy to detect connectivity failures. From the tests I've done, status code can be -1, 2 and very rarely 408.
However -1 is the generic status code we use for an unknown failure so we might end up using hysteresis with non connectivity issues.

PR #617, opened 2022-03-03
Filter afisafi

ddutt:
This has 3 patches, unrelated to each other, but here so all tests pass again.

* The addition of the filter keyword afiSafi to BGP and topology. 
* The fixing of the ospf help message
* The fixing of the sqpoller describe failure because it thinks the user has provided an argument.

### DIFF
--- a/suzieq/engines/pandas/topology.py
+++ b/suzieq/engines/pandas/topology.py
@@ -82,6 +82,7 @@ class TopologyObj(SqPandasEngine):
         area = kwargs.pop('area', '')
         vrf = kwargs.pop('vrf', '')
         addnl_fields = kwargs.pop('addnl_fields', [])
+        afi_safi = kwargs.pop('afiSafi', '')
 
         self._init_dfs(self._namespaces)
         self.lsdb = pd.DataFrame()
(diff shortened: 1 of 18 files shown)


LucaNicosia (approved):

claudious96 (approved):

suzieq/engines/pandas/topology.py
```diff
@@ -95,7 +96,11 @@ class Services:
         if peerHostname and 'peerHostname' not in fields:
             addnl_fields.append('peerHostname')
 
-        if 'asn' in columns or asn:
+        if (asn or afi_safi) and area:
```
claudiolor, line 99: Why not having this check in the `validate_get_input()` function of the sqobject?

claudiolor (approved):

suzieq/engines/pandas/topology.py
ddutt: Ideally all these checks should be done in sqobj. What we'd need to provide are mutually exclusive arguments and have those checked to make the checks table-independent. But maybe its OK to start a version now in topology alone and let it grow to include others. OTOH, we also have to check a bunch of other things such as what the type of each input is and reject it if its of a different type. So, I'll create a separate issue to handle all input validation in sqobj and continue with this patch.

PR #624, opened 2022-03-05
Consistency fixes

ddutt:


suzieq/engines/pandas/engineobj.py
```diff
@@ -178,12 +181,8 @@ def get_valid_df(self, table: str, **kwargs) -> pd.DataFrame:
         for fld in key_fields:
             if fld not in fields+addnl_fields:
                 addnl_fields.insert(0, fld)
-                drop_cols.append(fld)
 
-        for f in addnl_fields:
-            if f not in fields:
-                # timestamp is always the last field
-                fields.insert(-1, f)
+        getcols = set(fields+addnl_fields)
```
claudious96, line 185: This should be a list
```suggestion
        getcols = list(set(fields+addnl_fields))
```

suzieq/engines/pandas/engineobj.py
```diff
@@ -276,6 +275,9 @@ def get(self, **kwargs) -> pd.DataFrame:
         if not self.iobj.table:
             raise NotImplementedError
 
+        # Make sure we fetch all columns necessary including those that are
+        # needed to satisfy the filters
+
```
claudious96, line 280: Is this a TODO?

suzieq/engines/pandas/routes.py
```diff
@@ -49,8 +44,13 @@ def get(self, **kwargs):
         addnl_fields = kwargs.pop('addnl_fields', [])
         user_query = kwargs.pop('query_str', '')
 
-        columns = kwargs.get('columns', ['default'])
-        addnl_fields, drop_cols = self._cons_addnl_fields(
+        columns = kwargs.pop('columns', ['default'])
+        fields = self.schema.get_display_fields(columns)
+        if addnl_fields:
+            # if user requests additional fields, honor returning them
+            fields += [x for x in addnl_fields if x not in fields]
```
claudious96, line 51: Aren't we using `addnl_fields` only internally? We should only return the columns specified with `columns`.
Otherwise what is the purpose of `addnl_fields`?

suzieq/engines/pandas/address.py
```diff
@@ -33,17 +33,18 @@ def get(self, **kwargs) -> pd.DataFrame:
 
         addr = kwargs.pop("address", [])
         prefix = kwargs.pop("prefix", [])
-        columns = kwargs.get("columns", [])
+        columns = kwargs.pop("columns", [])
         ipvers = kwargs.pop("ipvers", "")
         user_query = kwargs.pop("query_str", "")
+        addnl_fields = kwargs.pop('addnl_fields', [])
```
LucaNicosia, line 39: Here we lose 'master' as default addnl_field. Is it calculated or a mistake?

suzieq/engines/pandas/engineobj.py
ddutt: I know, but does it matter?

suzieq/engines/pandas/engineobj.py
ddutt: No, just a comment.

suzieq/engines/pandas/routes.py
ddutt: Sometimes more complex tables such as topology and path uses them, and they need them. Why? Because its easier to say default + additional fields. We'll add the "+" and "-" notation later so that we can do things like columns='+macaddr' meaning the default columns plus macaddr, later. At that point, this remark will be true.

suzieq/engines/pandas/address.py
ddutt: Nope, master is an alias for vrf. If you notice, the address schema has no master.

claudiolor (approved):

suzieq/engines/pandas/routes.py
claudious96: We don't to that for most of the tables. Have a look at Device, Bgp, EvpnVni just to name the first that I checked. We always return `df[fields]` where fields is not extended with `addnl_fields`.

suzieq/engines/pandas/engineobj.py
claudious96: Ok, but I don't see the code implementing that comment. Am I missing something?

suzieq/engines/pandas/routes.py
ddutt: Those tables are not called by the other more complex tables.

ddutt: OK, routes is the sole one who does this @claudious96 , and so I'll fix it correctly and remove this.

ddutt: Interfaces does the same thing as routes, and both are invoked by more complex tables such as network, topology and path. One reason only interfaces and routes do it is because in the other cases, either the default list is enough or the list of fields required is small enough that we can specify the set.

suzieq/engines/pandas/engineobj.py
```diff
@@ -156,7 +158,10 @@ def get_valid_df(self, table: str, **kwargs) -> pd.DataFrame:
         # We cannot use sets because we want to preserve the order of fields
         filter_fields = [x for x in self.iobj.get_filter_fields
                          if x not in addnl_fields]
-        fields += [x for x in filter_fields+addnl_fields if x not in fields]
+        user_query_cols = [x for x in self._get_user_query_cols(query_str)
```
claudiolor, line 161: Why we inject the query string fields both in `get()` and `get_valid_df()` wouldn't doing it only in the latter be enough?

ddutt: Good point, fixing.

suzieq/engines/pandas/engineobj.py
ddutt: Dead comment :) Removed.

PR #626, opened 2022-03-06
Interfaces: fix vlan filter

claudious96:


suzieq/engines/pandas/interfaces.py
```diff
@@ -532,18 +532,16 @@ def _add_portmode(self, df: pd.DataFrame):
...
+        if not pm_df.empty:
+            df = df.merge(pm_df, how='left', on=[
+                'namespace', 'hostname', 'ifname'],
+                suffixes=['', '_y'])
+
+        df['portmode'] = np.where(df.ipAddressList.str.len() == 0,
+                                  'unknown',
+                                  'routed')
+        df['portmode'] = np.where(df.ip6AddressList.str.len() != 0,
+                                  'routed', df.portmode)
```
ddutt, line 543: I had made fixing of vlan to 0 as well. Is that done or was it unnecessary?

claudious96: It was a couple of rows below, it caused this error:
```
0  ERROR: The truth value of a Series is ambiguous. Use a.empty, a.bool(), a.item(), a.any() or a.all().
```
In the data I have there it didn't occur that any vlan had a NaN value. If you think it's possible we can do a `fillna(0)` before returning the df.

LucaNicosia (approved):

claudiolor (approved):

PR #638, opened 2022-03-18
Iosxe old dev

ddutt:
This is to support older IOS/XE devices. This support includes:

* Adding support for privilege escalation via enable, if specified (use enable-password to specify the password for enable, same choices as for password)
* Adding support for creating a single persistent interactive SSH session and using that for IOS/XE 
* Adding a new user-specified option to allow dealing with slow hosts by providing a delay of 5s between commands
* Update asyncssh library to be 2.9.0 to support interactive session


suzieq/poller/worker/nodes/node.py
```diff
@@ -1118,10 +1121,8 @@ async def _init_ssh(self, init_boot_time=True, _: bool = True) -> None:
...
         '''
         backoff_period = 1
-        while True:
+        while not self._conn:
             await super()._init_ssh(init_boot_time=False, rel_lock=False)
-            if self._conn:
-                break
 
             await asyncio.sleep(backoff_period)
             backoff_period *= 2
```
LucaNicosia, line 1130: Removing the `break` we are always waiting the backoff_period. Is it what we want? If yes, why?

suzieq/poller/worker/nodes/node.py
```diff
@@ -1187,6 +1188,93 @@ async def rest_gather(self, service_callback, cmd_list, cb_token,
...
+
+        while not self._conn:
+            await super()._init_ssh(init_boot_time=False, rel_lock=False)
+            if not self._conn:
+                await asyncio.sleep(backoff_period)
+                backoff_period *= 2
+                backoff_period = min(backoff_period, 120)
+                if rel_lock:
+                    self.ssh_ready.release()
+                return
```
LucaNicosia, line 1213: I'm not sure I understood what you are doing here. Why are you setting the backoff_period if you are going to return?That value will be lost. 
Moreover, that return makes the `while not self._conn` useless because you will never trigger it more than 1 time.

Another comment is related to the fact that you are releasing the lock after the backoff_period. What is the reason to keep the lock while it is sleeping? To slow down the commands?

suzieq/poller/worker/nodes/node.py
```diff
@@ -1187,6 +1188,93 @@ async def rest_gather(self, service_callback, cmd_list, cb_token,
...
+                self._stdin, self._stdout, self._stderr = \
+                    await self._conn.open_session(term_type='xterm')
+                self.logger.info(
+                    f'Persistent SSH created for {self.hostname}')
+
+                output = await self.wait_for_prompt()
+                if output.strip().endswith('>'):
+                    # We're not in enable mode and cannot function
+                    self.logger.warning(
+                        'Privilege escalation required. Trying')
```
LucaNicosia, line 1226: Add self.hostname in logs for better undestanding

suzieq/poller/worker/nodes/node.py
```diff
@@ -1187,6 +1188,93 @@ async def rest_gather(self, service_callback, cmd_list, cb_token,
...
+                    if self.enable_password:
+                        self._stdin.write(self.enable_password + '\n')
+                    else:
+                        self._stdin.write(self.password + '\n')
+
+                    output = await self.wait_for_prompt()
+                    if (output in ['timeout', 'Password:'] or
+                            output.strip().endswith('>')):
+                        self.logger.error(
+                            'Invalid enable password, commands may fail')
```
LucaNicosia, line 1238: As above

suzieq/poller/worker/nodes/node.py
```diff
@@ -1187,6 +1188,93 @@ async def rest_gather(self, service_callback, cmd_list, cb_token,
...
+                        await self._conn.terminate()
+                        self._conn = None
+                        if rel_lock:
+                            self.ssh_ready.release()
+                        return
+                    else:
+                        self.logger.warning('Privilege escalation succeeded')
```
LucaNicosia, line 1245: As above

suzieq/poller/worker/nodes/node.py
ddutt: Mistake. FIxed

suzieq/poller/worker/nodes/node.py
ddutt: Ugh. Fixing.

suzieq/poller/worker/nodes/node.py
ddutt: Done

suzieq/poller/worker/nodes/node.py
ddutt: Done

suzieq/poller/worker/nodes/node.py
ddutt: Done

PR #644, opened 2022-03-19
Misc fixes

ddutt:
A bunch of fixes, some ports, some real.

### DIFF
--- a/suzieq/poller/worker/services/bgp.py
+++ b/suzieq/poller/worker/services/bgp.py
@@ -96,10 +96,7 @@ class BgpService(Service):
                         new_entry['communityTypes'] = ['standard', 'extended']
                 else:
                     new_entry['communityTypes'] = ['standard']
-                if entry.get('hopsMax', '').isnumeric():
-                    # Older versions of EOS don't provide this field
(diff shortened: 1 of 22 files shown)


claudiolor (approved):

suzieq/poller/worker/services/bgp.py
```diff
@@ -96,10 +96,7 @@ def _clean_eos_data(self, processed_data, raw_data):
                         new_entry['communityTypes'] = ['standard', 'extended']
                 else:
                     new_entry['communityTypes'] = ['standard']
-                if entry.get('hopsMax', '').isnumeric():
-                    # Older versions of EOS don't provide this field
-                    new_entry['hopsMax'] = int(entry.get('hopsMax', 255))-255
```
claudious96, line 101: We don't need to do this anymore, right?

claudious96 (approved):

suzieq/poller/worker/services/bgp.py
ddutt: Correct, this was the only bug that needed to be fixed for capturing the TTL info correctly for multihop eBGP

PR #645, opened 2022-03-20
Evpn improv

ddutt:
A few EVPN improvements that had been squashed and not implemented.

### DIFF
--- a/suzieq/engines/pandas/evpnVni.py
+++ b/suzieq/engines/pandas/evpnVni.py
@@ -107,6 +107,11 @@ class EvpnvniObj(SqPandasEngine):
                         .rename(columns={'macaddr': 'numMacs'}) \
                         .fillna({'numMacs': 0})
 
+        for x in ['vlan', 'vni']:
+            if x in df.columns:
+                df[x] = np.where(df[x].isnull(), 0,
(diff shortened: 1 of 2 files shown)


suzieq/engines/pandas/evpnVni.py
```diff
@@ -107,6 +107,10 @@ def get(self, **kwargs) -> pd.DataFrame:
                         .rename(columns={'macaddr': 'numMacs'}) \
                         .fillna({'numMacs': 0})
 
+        for x in ['vlan', 'vni']:
+            if x in df.columns:
+                df[x] = df[x].astype(int)
```
claudious96, line 112: Is it possible that the column `vni` has `NaN` values? If yes it could crash here.

claudiolor (approved):

LucaNicosia (approved):

suzieq/engines/pandas/evpnVni.py
ddutt: OK, will address that

ddutt: Made use of np.where to address this

claudious96 (approved):

PR #646, opened 2022-03-21
Misc iosxe fixes

ddutt:
Found a bunch of IOSXE issues when debugging with a user.

### DIFF
--- a/suzieq/engines/pandas/engineobj.py
+++ b/suzieq/engines/pandas/engineobj.py
@@ -2,6 +2,7 @@ from typing import List
 import re
 from ipaddress import ip_address, ip_network
 
+from pytz import all_timezones
 import dateparser
 import numpy as np
 import pandas as pd
@@ -201,7 +202,10 @@ class SqPandasEngine(SqEngineObj):
         getcols = list(set(fields+addnl_fields))
 
         tz = self.ctxt.cfg.get('analyzer', {}).get('timezone', 'UTC')
-        settings = {'TIMEZONE': tz, 'TO_TIMEZONE': 'UTC'}
+        if tz not in all_timezones:
+            raise AttributeError(f'Invalid Timezone {tz} specified in config')
+
+        settings = {'TIMEZONE': tz}
 
         if self.iobj.start_time:
             try:
--- a/suzieq/engines/pandas/engineobj.py
+++ b/suzieq/engines/pandas/engineobj.py
@@ -202,9 +202,6 @@ class SqPandasEngine(SqEngineObj):
         getcols = list(set(fields+addnl_fields))
 
         tz = self.ctxt.cfg.get('analyzer', {}).get('timezone', 'UTC')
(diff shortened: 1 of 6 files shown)


suzieq/engines/pandas/engineobj.py
```diff
@@ -201,7 +202,10 @@ def get_valid_df(self, **kwargs) -> pd.DataFrame:
         getcols = list(set(fields+addnl_fields))
 
         tz = self.ctxt.cfg.get('analyzer', {}).get('timezone', 'UTC')
-        settings = {'TIMEZONE': tz, 'TO_TIMEZONE': 'UTC'}
+        if tz not in all_timezones:
+            raise AttributeError(f'Invalid Timezone {tz} specified in config')
+
+        settings = {'TIMEZONE': tz}
```
LucaNicosia, line 208: Isn't it better to have this check during the load of the sq_confing in the sqobject rather than in the engine?

suzieq/poller/worker/services/interfaces.py
```diff
@@ -742,6 +743,15 @@ def _clean_iosxr_data(self, processed_data, raw_data):
...
                 continue
 
+            if not 'type' in entry:
+                # We have some weird outputs when some ephemeral interfaces
+                # such as NDE_0 popup between the commands we run causing us
+                # to not have full data about this interface. Drop such
+                # interfaces. If they're not ephemeral, we'll get them in the
+                # next run
+                drop_indices.append(i)
+                return
```
claudious96, line 753: Pylint is complaining here, we should return an expression. We might:
1. return an empty dictionary
2. use `# pylint: disable=inconsistent-return-statements`

suzieq/poller/worker/services/interfaces.py
```diff
@@ -742,6 +743,15 @@ def _clean_iosxr_data(self, processed_data, raw_data):
                 entry['speed'] = NO_SPEED
                 continue
 
+            if not 'type' in entry:
```
claudious96, line 746:
```suggestion
            if 'type' not in entry:
```

claudious96 (changes requested):

suzieq/poller/worker/services/interfaces.py
claudiolor:
```suggestion
                continue
```
Shouldn't this return be a `continue`? In this way we discard all the interfaces

claudiolor (changes requested):

suzieq/poller/worker/services/interfaces.py
ddutt: YES!! I found and fixed it already

suzieq/engines/pandas/engineobj.py
ddutt: Its already there. Removed it. I found it crasjed on a mistake. Maybe its something else. need to validate

PR #647, opened 2022-03-21
Gui infinite rows

ddutt:
Adding start and end rows to the GUI Xplore page.

### DIFF
--- a/suzieq/gui/stlit/xplore.py
+++ b/suzieq/gui/stlit/xplore.py
@@ -12,6 +12,9 @@ from suzieq.gui.stlit.guiutils import (SUZIEQ_COLOR, SuzieqMainPages,
 from suzieq.gui.stlit.pagecls import SqGuiPage
 from suzieq.sqobjects import get_sqobject, get_tables
 
+DEF_MAX_ROW = 10000
+MAX_ROW_WARN = 50000
+
 
 @dataclass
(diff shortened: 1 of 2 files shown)


ddutt:
Ignore package updates and such

LucaNicosia (commented):
You need to remove the first 4 commits which are in the other PR. Apart from that, I don't see any problem in the gui

claudious96 (approved):

suzieq/gui/stlit/xplore.py
```diff
@@ -349,6 +384,17 @@ def _draw_aggrid_df(self, df) -> AgGrid:
...
+        start_row = self._state.start_row
+        end_row = self._state.end_row
+        if end_row > df.shape[0]:
+            end_row = df.shape[0]
+
+        if end_row - start_row > MAX_ROW_WARN:
+            st.warning('Printing too many rows can fail or be very slow')
+
+        if (end_row - start_row) < df.shape[0]:
+            st.info(f'Displaying rows {start_row} to {end_row}')
```
claudiolor, line 396:
```suggestion
            st.info(f'Displaying rows {start_row} to {end_row} out of {df.shape[0]}')
```
Knowing the size of the table might be useful when users select the rows to show

claudiolor (approved):

PR #652, opened 2022-03-22
More misc fixes

ddutt:
A bunch of minor issues, and one major issue that I found regarding filters. "!" operator wasn't working and we had no tests to catch it. Added additional tests as well. 

When you say hostname='leaf01 leaf02' we interpret that as leaf01 or leaf02. However, when you do hostname='!leaf01 !leaf02' we mean NOT leaf01 AND NOT leaf02. Handling this.

This also includes not doing hostname.unique().tolist() in cases where it breaks. Claudio Usai, I needed this done and found the other issues in the process, and so finished it here.


claudious96 (approved):

claudious96:
Some tests using two ! in the hostname filter are failing, just need to update the data for them.

suzieq/engines/pandas/lldp.py
```diff
@@ -109,17 +112,17 @@ def _resolve_to_bond(self, df: pd.DataFrame) -> pd.DataFrame:
...
             df: The original LLDP dataframe
+            hostname: The hostname selection passed by user
 
         Returns:
             The LLDP dataframe with the appropriate substitution
         """
 
-        iflist = df.ifname.unique().tolist() + \
-            df.peerIfname.unique().tolist()
+        df.peerIfname.unique().tolist()
```
LucaNicosia, line 121: What is this line for? The result is not assigned to anything

suzieq/sqobjects/basicobj.py
```diff
@@ -56,7 +56,7 @@ def __init__(self, engine_name: str = '',
 
         if engine_name and engine_name != '':
             self.engine = get_sqengine(engine_name, self._table)(self)
-        elif self.ctxt.engine:
+        elif self.ctxt.cfg.get('ux', {}).get('engine', 'pandas'):
             self.engine = get_sqengine(self.ctxt.engine, self._table)(self)
```
claudiolor, line 60: I think here you are always using pandas, even if in the config there is `rest`. That's because you are passing `self.ctxt.engine`, which is always set to pandas by the[ post_init of the context](https://github.com/netenglabs/suzieq/blob/develop/suzieq/shared/context.py#L30), unless explicitly set, and this only set by the CLI. Which means that in all the other cases we still go on using pandas even though the engine in the conf file is rest.
However I'm not sure this solution will work, even if you pass the right engine, then, you won't get all the parameters needed to connect to the rest server (address, port and api key), so I think we should move what is [here](https://github.com/netenglabs/suzieq/blob/develop/suzieq/cli/sq_nubia_context.py#L41) in the post_init of the context. But this is not enough, because we need to prevent the rest server from using the rest engine.

claudiolor (changes requested):

suzieq/sqobjects/basicobj.py
ddutt: Sigh. I could've sworn I tested the code. Anyways, you're right. Fixed it.

PR #653, opened 2022-03-22
Transport fix

LucaNicosia:
Now, if an invalid transport is passed, an error is raised to the user

### DIFF
--- a/suzieq/shared/utils.py
+++ b/suzieq/shared/utils.py
@@ -1,28 +1,27 @@
+import errno
+import fcntl
+import getpass
+import json
+import logging
 import os
 import re
 import sys
-from typing import Dict, List
-import logging
-from logging.handlers import RotatingFileHandler
 from datetime import datetime
-import fcntl
+from enum import Enum
 from importlib.util import find_spec
-from itertools import groupby
 from ipaddress import ip_network
-import errno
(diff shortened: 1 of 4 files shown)


claudious96 (approved):

claudiolor (approved):

suzieq/shared/utils.py
```diff
@@ -36,10 +35,15 @@
...
 NO_SPEED = 0
 MISSING_SPEED_IF_TYPES = ['ethernet', 'bond', 'bond_slave']
-SUPPORTED_POLLER_TRANSPORTS = ['ssh', 'https']
 SUPPORTED_ENGINES = ['pandas', 'rest']
 
 
+class PollerTransport(str, Enum):
+    """Supported poller transoport enum"""
+    ssh = 'ssh'
+    https = 'https'
```
ddutt, line 44: You need to support both http and https as transports.

ddutt (changes requested):

suzieq/shared/utils.py
LucaNicosia: It's not what we are doing in the node as you can see [here](https://github.com/netenglabs/suzieq/blob/master/suzieq/poller/worker/nodes/node.py#L756)

PR #662, opened 2022-03-28
IOSXE: Various fixes for hnags/timeouts

ddutt:
This patch contains a set of fixes when privilege escalation
is required. We were resulting in irrecoverable hangs, timeouts, and so
on. They were caused by not handling erroneous returns correctly in some
cases.

This doesn't handle the case of avoiding lockout by trying the wrong
password too many times because we initiate connection over and over
again.

We also return a consistent error code in case of timeouts now.

### DIFF
--- a/suzieq/poller/worker/nodes/node.py
+++ b/suzieq/poller/worker/nodes/node.py
@@ -118,6 +118,7 @@ class Node:
         self.sigend = False
         self.api_key = None
         self._stdin = self._stdout = None
+        self._retry = True      # set to False if authentication fails
 
         self.address = kwargs["address"]
         self.hostname = kwargs["address"]  # default till we get hostname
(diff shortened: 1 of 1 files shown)


suzieq/poller/worker/nodes/node.py
```diff
@@ -1318,12 +1324,14 @@ async def ssh_gather(self, service_callback, cmd_list, cb_token, oformat,
         result = []
         if cmd_list is None:
             await service_callback({}, cb_token)
+            return
 
-        if not self._stdin:
+        if not self._conn or not self._stdin:
             await self._init_ssh()
```
LucaNicosia, line 1330: Isn't it better to check the connection before calling the service callback?

claudious96 (approved):

claudiolor (approved):

claudious96 (approved):

PR #666, opened 2022-03-29
Last17 fixes

ddutt:


suzieq/poller/worker/nodes/node.py
```diff
@@ -1266,7 +1266,9 @@ async def _init_ssh(self, init_boot_time=True,
                         self.logger.info(
                             'Privilege escalation succeeded for '
                             f'{self.hostname}')
-            except Exception:
+            except Exception as e:
+                self.logger.error('Unable to create persistent SSH session'
+                                  f' for {self.hostname} due to {str(e)}')
```
claudious96, line 1271: Shouldn't we also set `self._current_exception`?

Dockerfile-demo
```diff
@@ -1,6 +1,8 @@
 FROM netenglabs/suzieq:latest
 
-COPY ./parquet /suzieq/parquet
+ARG username=suzieq
+
+COPY ./tests/data/parquet /home/$username/parquet
```
claudiolor, line 5: I think this Dockerfile doesn't need any modification, having the parquet data in the user directory is wrong, since the config file points to the /suzieq directory

claudiolor (changes requested):

LucaNicosia (approved):

Dockerfile-demo
ddutt: I have this since I don't expect the user to do anything in the demo.

suzieq/poller/worker/nodes/node.py
ddutt: Yes, will fix

PR #683, opened 2022-03-31
Dockerfile updates

### LINKED ISSUE #678: Docker file permission issues with volumes
**Describe the bug**
Because of the move to a non-root user, you need to be more mindful of Docker directory permissions with volumes.  Users are hitting this issue quite a bit

**To Reproduce**
1. Start a new container with a volume for parquet
2. See errors related to file permissions because the directory is set with root ownership

**Attach**
Coming

 **Expected behavior**
Volume to use the suzieq user in the container

**Screenshots**

**Additional context**

ryanmerolle:
## Description

Fixes #678

- Addresses Dockerfile RUN consolidation per best practices and to shrink container size
- Fixes docker container parquet volume/directory permission issues due to root ownership being applied
- Adds a .dockerignore file per best practices (need to make sure nothing needed in the docker containers is ignored)
- Moves to explicit `python3 -m pip install` for pip installs

## Type of change

- Optimization of Dockerfiles

## Comments

The item we should double check is that we are not ignoring files that need to be in the docker image


anubisg1:
I would replace useradd with adduser

That would negate the whole need to chown the home folder or any subfile or subdirectory created in that folder.

Furthermore ~/parquet doesn't need to be created because it would be created at runtime by the suzieq user with the correct permissions or should be mounted properly by the user as a volume.

In kubernetes for example you'd use "fsgroup" under the security context

Dockerfile
```diff
@@ -10,11 +10,15 @@ RUN useradd $username -u $user_id --create-home --user-group
...
-RUN pip install /tmp/suzieq-$version-py3-none-any.whl
-RUN rm /tmp/suzieq-$version-py3-none-any.whl
 COPY --chown=$username suzieq/config/etc/suzieq-cfg.yml /home/$username/.suzieq/suzieq-cfg.yml
-RUN sed -i 's/127.0.0.1/0.0.0.0/' /home/$username/.suzieq/suzieq-cfg.yml
+
+RUN python3 -m pip install --upgrade pip && \
+    python3 -m pip install /tmp/suzieq-$version-py3-none-any.whl && \
+    rm /tmp/suzieq-$version-py3-none-any.whl && \
+    mkdir /home/$username/parquet && \ 
+    sed -i 's/127.0.0.1/0.0.0.0/' /home/$username/.suzieq/suzieq-cfg.yml
```
claudiolor, line 19:
```suggestion
RUN python3 -m pip install --upgrade pip

COPY --chown=$username suzieq/config/etc/suzieq-cfg.yml /home/$username/.suzieq/suzieq-cfg.yml

RUN mkdir /home/$username/parquet && \
    chown $user_id:$user_id /home/$username/parquet && \
    sed -i 's/127.0.0.1/0.0.0.0/' /home/$username/.suzieq/suzieq-cfg.yml

COPY ./dist/suzieq-$version-py3-none-any.whl  /tmp/

RUN python3 -m pip install /tmp/suzieq-$version-py3-none-any.whl && \
    rm /tmp/suzieq-$version-py3-none-any.whl
```
I agree about reducing the number of layer, but I would split the RUNs in order to use caching, so reducing build time and memory usage. For example, we don't need to update pip again if the wheel changes.

Additionally probably you forgot to chown parquet directory.

Dockerfile
```diff
@@ -10,11 +10,15 @@ RUN useradd $username -u $user_id --create-home --user-group
 
 COPY ./dist/suzieq-$version-py3-none-any.whl  /tmp/
```
claudiolor, line 11:
```suggestion
```
If the wheel changes we invalidate all the following layers, better to put it some lines after (already included in the previous comment)

Dockerfile
```diff
@@ -10,11 +10,15 @@ RUN useradd $username -u $user_id --create-home --user-group
...
 
-RUN pip install --upgrade pip
-RUN pip install /tmp/suzieq-$version-py3-none-any.whl
-RUN rm /tmp/suzieq-$version-py3-none-any.whl
 COPY --chown=$username suzieq/config/etc/suzieq-cfg.yml /home/$username/.suzieq/suzieq-cfg.yml
```
claudiolor, line 13:
```suggestion
```
Same here

Dockerfile
```diff
@@ -10,11 +10,15 @@ RUN useradd $username -u $user_id --create-home --user-group
...
-RUN sed -i 's/127.0.0.1/0.0.0.0/' /home/$username/.suzieq/suzieq-cfg.yml
+
+RUN python3 -m pip install --upgrade pip && \
```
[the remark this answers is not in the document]
ryanmerolle: Great question.  It was [officially recommended in the python docs](https://docs.python.org/3/installing/index.html#basic-usage) as the way to install back in 2014.

There are a number of articles about it:

- https://stackoverflow.com/questions/25749621/whats-the-difference-between-pip-install-and-python-m-pip-install
- https://www.reddit.com/r/learnpython/comments/mqsp3u/python_m_pip_install_vs_pip_install/
- https://snarky.ca/why-you-should-use-python-m-pip/

Ultimately it will make no difference here.  Its just habit, and I am happy to revert.

ryanmerolle:
> I would replace useradd with adduser
> 
> That would negate the whole need to chown the home folder or any subfile or subdirectory created in that folder.
> 
> Furthermore ~/parquet doesn't need to be created because it would be created at runtime by the suzieq user with the correct permissions or should be mounted properly by the user as a volume.
> 
> In kubernetes for example you'd use "fsgroup" under the security context

I still had an issue with `adduser`  maybe I was testing incorrectly.  Any suggestions or tests would be much appreciated if there is a better approach

Dockerfile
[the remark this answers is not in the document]
claudiolor: As far as I understood the main reason for using `python3 -m pip` is to have more control on the version of python we are using and to have a command that works even when the environment is not correctly configured. But since we are in a container, which is a controlled environment, where we have a single version of python I would go with just `pip`, it's shorted and if we have a problem with it, the problem is not using just `pip` but the way we are building the container.

claudiolor:
All the rest looks great to me, thank you!

ryanmerolle:
With the PR now merged, can we please get new images built and pushed up?

PR #688, opened 2022-04-04
Fix load_sq_config calls in worker tests

claudiolor:
## Description
This PR fixes some bugs in tests causing the configuration file in the default location to be used for testing:
- Some of the unit tests called the `load_sq_config()` with the wrong arguments, so these tests used the configuration file in the default location, instead of the generated dummy configuration file
- Some of the poller controller tests used the configuration file in the default location

## Type of change


tests/unit/poller/controller/test_controller.py
```diff
@@ -3,6 +3,7 @@
 import os
 import signal
 from tempfile import NamedTemporaryFile
+from tkinter import N
```
claudious96, line 6: I think this was not added on purpose

PR #692, opened 2022-04-09
DB: Fix time handling

ddutt:
The DB is optimized to provide constant query times independent of the
time window specified. The code to read the set of files given the
variety of user inputs was buggy and read a lot more files than
necessary. This reduced the time to respond to user queries. This patch
addresses this.

### DIFF
--- a/suzieq/db/parquet/parquetdb.py
+++ b/suzieq/db/parquet/parquetdb.py
@@ -509,17 +509,6 @@ class SqParquetDB(SqDB):
 
         folder = self._get_table_directory(table_name, True)
 
-        # pylint: disable=simplifiable-if-statement
-        if not (start_time or end_time) and (view == "all"):
-            # Enforcing the logic we have: if both start_time & end_time
-            # are given, return all files since the model is that the user is
-            # expecting to see all changes in the time window. Otherwise, user
-            # expecting to see only the latest before an end_time OR after a
-            # start_time.
-            all_files = True
-        else:
-            all_files = False
-
         # We need to iterate otherwise the differing schema from different dirs
         # causes the read to abort.
         dirs = Path(folder)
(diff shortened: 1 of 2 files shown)


suzieq/db/parquet/parquetdb.py
```diff
@@ -509,16 +509,8 @@ def _get_cp_dataset(self, table_name: str, need_sqvers: bool,
...
-            # Enforcing the logic we have: if both start_time & end_time
-            # are given, return all files since the model is that the user is
-            # expecting to see all changes in the time window. Otherwise, user
-            # expecting to see only the latest before an end_time OR after a
-            # start_time.
-            all_files = True
-        else:
-            all_files = False
+        # pylint: disable=consider-using-ternary
+        all_files = ((start_time and end_time) or (view == "all"))
```
LucaNicosia, line 513: all_files is true only for both start_time and end_time are set (or view == 'all'). Shouldn't we set all_files to False also if only one between start_time or end_time are specified?
Isn't it `all_files = (not (start_time or end_time) or (view == 'all'))` what we should do?

claudious96: I agree with Luca, with this change if we set both `start_time` and `end_time` we are going to read every file of the dataset.

ddutt: I agree, Ugh. I had the original code, but it'd read all files if we only gave end-time. I don't the logic is here that needs to be fixed.

ddutt: The correct logic I think is this in English; if view = all is specified, and there are no time parameters, just suck in all files. So, its (view == "all") and not (start_time or end_time)

claudious96 (approved):

claudiolor:
I know the PR is already merged, but I left the comments anyway, if needed we can open another PR to address these problems. I left a comment about view=all and only the end-time filter. Additionally, some comments about additional comments and variable rename, as I add some difficulties with understanding the code while doing the review

PR #693, opened 2022-04-10
Lib upd8 017

ddutt:
Updating dependency lib due to security requirement. Also fixing Dockerfile for the poller to log to stdout by default.

Ignore the first 2 commits, I'll rebase before merging to remove them.


Dockerfile
```diff
@@ -14,7 +14,8 @@ COPY --chown=$username suzieq/config/etc/suzieq-cfg.yml /home/$username/.suzieq/
 
 RUN mkdir /home/$username/parquet && \
     chown $user_id:$user_id /home/$username/parquet && \
-    sed -i 's/127.0.0.1/0.0.0.0/' /home/$username/.suzieq/suzieq-cfg.yml
+    sed -i 's/127.0.0.1/0.0.0.0/' /home/$username/.suzieq/suzieq-cfg.yml && \
+    sed -i 's/# log-stdout:/log-stdout:/' /home/$username/.suzieq/suzieq-cfg.yml
```
claudiolor, line 18: Why don't we just fix the configuration file?

ddutt: Because that'd be so easy, and why would we want that :(. Stupid me. Fixing.

PR #694, opened 2022-04-10
First cut adding active column when showing active column

ddutt:
This adds the active column to the output with the suggestion Claudio Lorina made (+/- in CLI). I'll fix the GUI after, and add some tests later tonight.

With this, we should be able to release 0.17.1

### DIFF
--- a/suzieq/cli/sqcmds/command.py
+++ b/suzieq/cli/sqcmds/command.py
@@ -5,6 +5,7 @@ import inspect
 from io import StringIO
 import shutil
 
+import numpy as np
 import pandas as pd
 from nubia import command, context
 
@@ -447,6 +448,8 @@ class SqCommand(SqPlugin):
             for row in df.itertuples():
                 self._pager_print(row.config)
         else:
+            if 'active' in df.columns:
+                df['active'] = np.where(df.active, '+', '-')
             if (not (self.start_time or self.end_time) and
                 'timestamp' not in self.columns and not dont_strip_cols and
                     'timestamp' in df.columns and 'timestamp' in cols):
--- a/suzieq/engines/pandas/engineobj.py
+++ b/suzieq/engines/pandas/engineobj.py
@@ -188,9 +188,7 @@ class SqPandasEngine(SqEngineObj):
         if 'timestamp' not in fields:
             fields.append('timestamp')
 
-        if 'active' not in fields+addnl_fields:
-            addnl_fields.append('active')
-
+        self._add_active_to_fields(view, fields, addnl_fields)
         # Order matters. Don't put this before the missing key fields insert
         for f in aug_fields:
             dep_fields = sch.get_parent_fields(f)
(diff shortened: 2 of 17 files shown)


suzieq/engines/pandas/evpnVni.py
```diff
@@ -22,6 +22,9 @@ def get(self, **kwargs) -> pd.DataFrame:
 
         addnl_fields = []
         fields = self.schema.get_display_fields(columns)
+        self._add_active_to_fields(kwargs.get('view', 'latest'), fields,
+                                   addnl_fields)
```
LucaNicosia, line 26: `view` is not a kwargs. It's stored inside self.iobj.view. `kwargs.get('view', 'latest')` returns always 'latest'.
My suggestion is to modify the function `_add_active_to_fields`.

suzieq/engines/pandas/engineobj.py
```diff
@@ -699,3 +697,29 @@ def _get_user_query_cols(self, query_str: str) -> List[str]:
...
+            None
+        """
+
+        if view == "all" or (self.iobj.start_time and self.iobj.end_time):
+            if 'active' not in fields:
+                fields.insert(0, 'active')
+        else:
+            addnl_fields.append('active')
+
+        return
```
LucaNicosia, line 725:
```suggestion
    def _add_active_to_fields(self, fields: List[str],
                              addnl_fields: List[str]) -> None:
        """Add 'active' field to returned cols
        Depending on the user specification, if we have to return all
        records within a time window or if view=all, we must also
        return the active field because it marks if the record was
        added or deleted
        Args:
            fields(List[str]): the columns to be returned so far
            addnl_fields(List[str]): the addnl fields to be retrieved so far
        Returns:
            None
        """

        if self.iobj.view == "all" or (self.iobj.start_time and self.iobj.end_time):
            if 'active' not in fields:
                fields.insert(0, 'active')
        else:
            addnl_fields.append('active')

        return
```

suzieq/cli/sqcmds/command.py
```diff
@@ -447,6 +448,9 @@ def _gen_output(self, df: pd.DataFrame, json_orient: str = "records",
             for row in df.itertuples():
                 self._pager_print(row.config)
         else:
+            if (self.start_time and self.end_time) or self.view == "all":
+                if 'active' in df.columns:
+                    df['active'] = np.where(df.active, '+', '-')
```
LucaNicosia, line 453: We should change this check in order to have consistency. With this code, if we do `interface show columns='namespace hostname active'` we see `active` columns with `False/True`. I think is better if we show also here `+/-`.
```suggestion
            if 'active' in columns:
                df['active'] = np.where(df.active, '+', '-')
```

LucaNicosia (changes requested):

suzieq/engines/pandas/evpnVni.py
ddutt: We had a discussion about this on Slack. We're going to leave it as is for now

suzieq/engines/pandas/engineobj.py
ddutt: Fixed, also checking that addnl_fields is never None, because we pass that in ospf. Safeguarding it as against None seemed better than passing [] in ospf.py.

claudiolor (changes requested):

suzieq/engines/pandas/network.py
```diff
@@ -266,21 +266,43 @@ def _find_addr_arp(self, addr: str, **kwargs) -> pd.DataFrame:
...
+
             ifdf = self._get_table_sqobj('interfaces') \
                 .get(namespace=[row.namespace], hostname=[row.hostname],
                      ifname=[oif])
 
             if not ifdf.empty:
                 if 'vlan' not in ifdf.type.unique():
                     # Routed interface
-                    result.append({
+                    if hasattr('active', row):
```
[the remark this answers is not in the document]
ddutt: Fixed

PR #697, opened 2022-04-12
Update docker docs

claudiolor:
## Description
This PR provide some changes on the documentation and on the docker-file:
- Update the mount point of the docker volume
- Update the example configuration file of the rest section of the documentation
- Set the official SuzieQ image as default

## Type of change

- Documentation update

### DIFF
--- a/docs/rest-server.md
+++ b/docs/rest-server.md
@@ -13,7 +13,7 @@ You must launch the Suzieq docker container as follows:
 ```
 This assumes that you're using port 8000 to connect to the REST server. If you wish to use a different port for the REST server, say 7000, you can launch it as ```docker run -itd -p 7000:8000 --name suzieq netenglabs/suzieq:latest```.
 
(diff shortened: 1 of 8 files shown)


docs/rest-server.md
```diff
@@ -68,17 +68,11 @@ openssl rand -hex 20
 
 ### Changing the Location And Name of the Key/Cert Files
 
-The default location for the key and certificate files needed for REST is ~/.suzieq, and the file names are key,pem and cert.pem. If you wish these to be different, you can specify the full path of each of the files (or any one) via the ```rest_keyfile``` and ```rest_certfile``` variables in the suzieq config file, typically at ~/.suzieq/suzieq-cfg.yml. Here is an example of a config file with these two parameters defined:
-```
+The default location for the key and certificate files needed for REST is `~/.suzieq`, and the file names are `key,pem` and `cert.pem`. If you wish these to be different, you can specify the full path of each of the files (or any one) via the ```rest_keyfile``` and ```rest_certfile``` variables under the `rest` section of the suzieq config file, typically at ~/.suzieq/suzieq-cfg.yml. Here is an example of a config file with these two parameters defined:
```
LucaNicosia, line 71: You wrote `key,pem` instead of `key.pem`

PR #716, opened 2022-04-28
Node: One more fixup after refactoring

ddutt:
We had the true/false signs mixed up in checking for a live
connection in case of IOS/IOSXE devices and that screwed up the
connectivity to those devices. Furthermore, if the user supplies a
device type, we need to switch to it first as the device might require
privilege escalation and if we don't switch first, we'll never manage to
connect to the device.

## Type of change

## New Behavior

When a device type is supplied for IOS and privilege escalation is required, we connect to the device and everything works.

...

When a device type is supplied for IOS and privilege escalation is required, we cannot communicate with the device.

...
## Double Check

- [ X] I have read the comments and followed the [CONTRIBUTING.md](https://github.com/netenglabs/suzieq/blob/master/CONTRIBUTING.md).
- [ X] I have explained my PR according to the information in the comments or in a linked issue.
- [ X] My PR source branch is created from the `develop` branch.
- [ X] My PR targets the `develop` branch.
- [ X] All my commits have `--signoff` applied

### DIFF
--- a/suzieq/poller/worker/nodes/node.py
+++ b/suzieq/poller/worker/nodes/node.py
@@ -159,28 +159,32 @@ class Node:
 
         self._init_service_queue()
 
+        devtype = kwargs.get("devtype", None)
+        if devtype:
+            self._set_devtype(devtype, '')
+            self.logger.warning(
+                f'{devtype} supplied for {self.address}:{self.port}')
+
         if self.transport == "ssh":
             self.port = self.port or 22
             await self._init_ssh(init_dev_data=False)
         elif self.transport == "https":
             self.port = self.port or 443
-
-        devtype = kwargs.get("devtype", None)
-        if devtype:
-            self._set_devtype(devtype, '')
-            if self.transport == 'https':
+            if self.devtype:
+                # Checking devtype to ensure we didn't get a REST transport
+                # without also providing the device type.
                 await self._init_rest()
-            self.logger.warning(
-                f'{devtype} supplied for {self.address}:{self.port}')
-        elif self.transport != 'ssh':
-            self.logger.error(
-                'devtype MUST be specified in inventory file if transport'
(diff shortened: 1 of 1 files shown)


suzieq/poller/worker/nodes/node.py
```diff
@@ -159,28 +159,30 @@ async def initialize(self, **kwargs) -> TNode:
...
             await self._init_ssh(init_dev_data=False)
         elif self.transport == "https":
             self.port = self.port or 443
-
-        devtype = kwargs.get("devtype", None)
-        if devtype:
-            self._set_devtype(devtype, '')
-            if self.transport == 'https':
+            if self.devtype:
                 await self._init_rest()
```
LucaNicosia, line 174: Please add a comment here where you explain why we should init rest if the devtype is set. It's not clear to me

suzieq/poller/worker/nodes/node.py
```diff
@@ -1454,7 +1456,7 @@ async def _init_ssh(self, init_dev_data=True,
 
                 output = await self.wait_for_prompt()
                 if output.strip().endswith('>'):
-                    if not self._handle_privilege_escalation():
+                    if await self._handle_privilege_escalation():
```
LucaNicosia, line 1459: Can we change this check to `if await self._handle_privilege_escalation() == -1`? It's easier to read

claudious96 (approved):

suzieq/poller/worker/nodes/node.py
ddutt: OK. We need to add it because the devtype MAY not have been specified by user but transport provided is REST

suzieq/poller/worker/nodes/node.py
ddutt: OK, I'm used to the Unix way of thinking :)

PR #721, opened 2022-05-03
Nxos7

ddutt:

## Description

Added support for older NXOS versions for all commands except EVPN/VXLAN.

## Type of change

## New Behavior

...

Older NXOS versions, older than NXOS 9.3 are now supported in SuzieQ.

## Double Check


suzieq/poller/worker/services/arpnd.py
```diff
@@ -64,6 +64,10 @@ def _clean_nxos_data(self, processed_data, _):
                 drop_indices.append(i)
                 continue
 
+            if 'state' not in entry:
+                # textfsm version handling
+                entry['state'] = 'reachable'
```
LucaNicosia, line 69: I saw we are doing this also for other nosses, so I'm not complaining. It's just a curiosity.
Why can we assume `state = reachable` if we do not get any state?

suzieq/engines/pandas/ospf.py
```diff
@@ -425,12 +425,15 @@ def _get_peernames(self, df: pd.DataFrame, columns: List[str],
         if newdf.empty:
             return df
 
-        lldp_df = self._get_table_sqobj('lldp').get(
+        addr_df = self._get_table_sqobj('address').get(
```
LucaNicosia, line 428: Why do we changed from lldp to address here?

suzieq/poller/worker/services/ospfIf.py
```diff
@@ -126,6 +126,8 @@ def _clean_nxos_data(self, processed_data, _):
 
                 for j, area in enumerate(entry['ifname']):
                     for ifentry in areas.get(area, []):
+                        if ifentry['vrf'] != entry['vrf']:
+                            continue
```
LucaNicosia, line 130: please add a comment explaining why you are continuing here

suzieq/poller/worker/services/interfaces.py
```diff
@@ -592,6 +592,11 @@ def fix_nxos_speed(entry):
                 drop_indices.append(entry_idx)
                 continue
 
+            if entry.get('_entryType', '') == "vrf":
+                entry['type'] = 'vrf'
+                entry['mtu'] = 1500
+                entry['vni'] = 0
```
LucaNicosia, line 598: Can you add a comment explaining why you are setting these values? If we will have problem with nxos, these lines will be very confusing to me during the debugging phase

suzieq/engines/pandas/ospf.py
```diff
@@ -425,12 +425,15 @@ def _get_peernames(self, df: pd.DataFrame, columns: List[str],
...
             namespace=kwargs.get('namespace', []),
-            hostname=kwargs.get('hostname', []),
-            use_bond='True')
+            hostname=kwargs.get('hostname', []))
+        if not addr_df.empty:
+            addr_df = addr_df.query('ipAddressList.str.len() != 0')
```
claudious96, line 432: Should we be using `ipAddress` (v4 and v6) instead of `ipAddressList` (v4 only)? I'm not sure anyone uses ipv6 for ospf but it's a low effort to support them both.

suzieq/config/textfsm_templates/nxos_show_clock.tfsm
```diff
@@ -0,0 +1,10 @@
+Value time ([0-9.:]+)
+Value timezone (\S+)
+Value ntpEnabled (\w+)
+
+Start
+  ^${time}\s+${timezone}\s+.*$$
+  ^Time\s+[sS]ource\s+is\s+${ntpEnabled}
+
+#13:56:23.593 UTC Mon Apr 18 2022
+#Time source is NTP
```
claudious96, line 10: This file is missing a newline at EOF, linting will fail.

suzieq/engines/pandas/ospf.py
ddutt: We need to handle non-p2p interfaces (think Vlan) which doesn't work with lldp.

suzieq/poller/worker/services/interfaces.py
ddutt: These values are not provided by NXOS, ever.

PR #726, opened 2022-05-09
Pipelining

ddutt:
**THIS IS A DRAFT**

This patch enables support for throttling the number of simultaneous SSH connections OR commands executed across all workers. We add a new keyword to the suzieq config file, max-cmd-pipeline, which needs to be a mutliple of the number of workers launched. Each worker gets an equal amount of credits from this total via the runtime cmdline arg, worker-cmd-pipeline. 

We use asyncio.Semaphore to implement the throttle. Every SSH connection setup and every command execution (except setting terminal length to 0) needs to acquire the semaphore before proceeding. When multiple commands are executed per call to ssh_gather, we reduce the parallelizing of this to a single command execution when throttling is enabled. In this case, since the commands are executed one after the other, a single lock taken before command call is sufficient.

Command executions release the lock during the call to the callback function called after the command is done. This means that we must not attempt to take more than one lock when multiple commands are supplied to ssh_gather. 

Checks:
* [X] Check if max-cmd-pipeline is a multiple of num of workers, and error out
* [X] Reduce cmd batch size to 1 when throttling is enabled
* [X] Release lock in init_ssh before invoking init_dev_data since that call involves calling ssh_gather
* [X] Take a lock after ssh connection is initialized in ssh_gather
* [X] Take a lock before invoking privilege escalation


suzieq/poller/worker/nodes/node.py
```diff
@@ -690,7 +718,7 @@ async def _local_gather(self, service_callback: Callable,
                 self.current_exception = e
                 result.append(self._create_error(cmd))
 
-        await service_callback(result, cb_token)
+        await self.exec_svc_callback(service_callback, result, cb_token)
```
claudiolor, line 721: Should the local gather release the semaphore? It never acquires it

claudiolor (changes requested):

suzieq/poller/worker/nodes/node.py
```diff
@@ -725,9 +753,14 @@ async def _ssh_gather(self, service_callback: Callable,
                         "Unable to connect to node %s cmd %s",
                         self.hostname, cmd)
                     result.append(self._create_error(cmd))
-                await service_callback(result, cb_token)
+                    await self.exec_svc_callback(service_callback, result,
```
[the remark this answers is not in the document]
ddutt: Fixed

suzieq/poller/worker/nodes/node.py
ddutt: Not for now

suzieq/poller/worker/nodes/node.py
```diff
@@ -1143,7 +1178,7 @@ async def _rest_gather(self, service_callback, cmd_list, cb_token,
                               "%s:%d due to %s", self.address, self.port,
                               e)
 
-        await service_callback(result, cb_token)
+        await self.exec_svc_callback(service_callback, result, cb_token)
```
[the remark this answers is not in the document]
ddutt: We should not do this for REST either.

suzieq/engines/pandas/ospf.py
```diff
@@ -282,24 +286,27 @@ def aver(self, **kwargs):
...
+            return self._create_aver_result([ok_df, ifdown_df], result)
 
         # We have the MTU mismatch which requires us to pull the MTU from the
         # interfaces table
         ifdf = self._get_table_sqobj('interfaces').get(
-            namespace=failed_df.namespace.unique().tolist(),
-            hostname=failed_df.hostname.unique().tolist(),
-            ifname=failed_df.ifname.unique().tolist(),
+            namespace=kwargs.get('namespace', []),
+            hostname=kwargs.get('hostname', []),
```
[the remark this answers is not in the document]
ddutt: Yes. I should also remove the others as the list can get pretty long in large installations.

suzieq/engines/pandas/ospf.py
```diff
@@ -374,6 +384,15 @@ def aver(self, **kwargs):
                 axis=1,
             )
 
+        # Now walk through and label all the failed sessions for
+        # which we have no reason
+        peer_df['assertReason'] += peer_df.apply(
+            lambda x: ["no peer found"]
```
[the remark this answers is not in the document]
ddutt: Will do, thanks for catching it

suzieq/engines/pandas/interfaces.py
```diff
@@ -585,13 +585,14 @@ def _add_portmode(self, df: pd.DataFrame, **kwargs):
         if not pm_df.empty:
             df = df.merge(pm_df, how='left', on=[
                 'namespace', 'hostname', 'ifname'],
-                suffixes=['', '_y'])
+                suffixes=['', '_y']) \
+                .fillna({'vlan_y': 0})
```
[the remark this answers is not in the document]
ddutt: Good point.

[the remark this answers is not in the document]
LucaNicosia: Here is what you do a few line below [line 594](https://github.com/netenglabs/suzieq/blob/pipelining/suzieq/engines/pandas/interfaces.py#L594)

```python
if 'vlan_y' in df.columns:
            df['vlan'] = np.where(df.vlan != 0, df.vlan,
                                  df.vlan_y)
```
So, if both vlan and vlan_y are 0, the value of `df['vlan']` will be vlan_y so 0 and we cannot know if this 0 is actually a 0 comining from the parquet data or by the fillna

suzieq/poller/worker/nodes/node.py
```diff
@@ -232,6 +241,32 @@ def is_connected(self):
...
+        if self._cmd_sem:
+            self.logger.info(
+                f'{self.transport}://{self.hostname}:{self.port}: Get lock')
+            await self._cmd_sem.acquire()
+            self.logger.info(
+                f'{self.transport}://{self.hostname}:{self.port}: Got lock')
+            yield
+            self.logger.info(
+                f'{self.transport}://{self.hostname}:{self.port}: Free lock')
+            self._cmd_sem.release()
```
claudiolor, line 266:
```suggestion
            try:
              yield
            finally:
              self.logger.info(
                  f'{self.transport}://{self.hostname}:{self.port}: Free lock')
              self._cmd_sem.release()
```
Using finally we release the semaphore even when an exception is raised inside the context

suzieq/poller/worker/nodes/node.py
```diff
@@ -989,21 +1029,21 @@ async def run(self):
                 try:
                     request = await self._service_queue.get()
                 except asyncio.CancelledError:
+                    self.logger.info('releasing sempahore')
```
claudiolor, line 1032:
```suggestion
```

suzieq/poller/worker/nodes/node.py
```diff
@@ -989,21 +1029,21 @@ async def run(self):
...
                     await self._terminate()
                     return
 
                 if request:
                     tasks.append(self._exec_service(
                         request[0], request[1], request[2]))
-                    self.logger.debug(
+                    self.logger.info(
```
claudiolor, line 1039: Why we are moving it from debug to info? I think debug was fine, it would create a lot of noise in info

claudiolor (changes requested):

suzieq/poller/controller/manager/static.py
```diff
@@ -41,6 +42,25 @@ def __init__(self, config_data: Dict = None, validate: bool = True):
...
+        # We need a pipeline thats at least as big as the number of workers
+        max_cmd_pipeline = config_data['config-dict'] \
+            .get('poller', {}) \
+            .get('max-cmd-pipeline', 0)
+        if ((max_cmd_pipeline != 0) and
+                (max_cmd_pipeline % self._workers_count != 0)):
+            logger.error(f'max-cmd-pipeline({max_cmd_pipeline}) has to be a '
+                         'multiple of the number of workers'
+                         f'({self._workers_count})')
+            sys.exit(1)
```
claudiolor, line 54: Please raise an exception instead:
```suggestion
           raise SqPollerConfError(f'max-cmd-pipeline({max_cmd_pipeline}) has to be a '
                         'multiple of the number of workers'
                         f'({self._workers_count})')
```

claudiolor (changes requested):

suzieq/poller/worker/inventory/inventory.py
```diff
@@ -55,7 +55,7 @@ def running_nodes(self) -> List[Coroutine]:
         """
         return self._node_tasks
 
-    async def build_inventory(self) -> Dict[str, Node]:
+    async def build_inventory(self, cmd_sem) -> Dict[str, Node]:
```
claudious96, line 58: Could you add the type hint here and in the other functions with the semaphore as arg?

suzieq/poller/worker/nodes/node.py
```diff
@@ -1587,42 +1631,43 @@ async def _ssh_gather(self, service_callback, cmd_list, cb_token, oformat,
...
-                    try:
-                        await self._close_connection()
-                        self.logger.debug(
-                            f"Closed conn successfully for {self.hostname}")
-                    except Exception as e1:
+        async with self.limit_pipeline():
+            for cmd in cmd_list:
+                try:
+                    if self.slow_host:
+                        await asyncio.sleep(IOS_SLEEP_BET_CMDS)
```
claudious96, line 1638: Wouldn't be better to do the sleep before acquiring the semaphore?

PR #728, opened 2022-05-11
Restore network commands for backward compatibility

LucaNicosia:
## Description

This PR restore `network show/summarize/unique/top` commands. They calls underneath the related `namespace show/summarize/unique/top` commands. 
A warning is showed inside the CLI suggesting the new command to use.
image

The REST server routes for the network commands are now marked as deprecated.

Also the sqobjects by default print on stdout a similar warning is displayed on the CLI. It can be ignored passing `ignore_warning=True` as function argument.
 ```python
network_summ_result = get_sqobject('network')(..).summarize(..., ignore_warning=True)
```

## Type of change

## New Behavior

Now it is possible to call again `network show/summarize/top/unique` commands with the only difference of a warning showed on the CLI explaining that the command is deprecated. The result is the same.

`network show/summarize/top/unique` commands were removed.

For backward compatibility, we will keep both commands. In the future probably these network commands will be removed.

## Double Check

### DIFF
--- a/suzieq/shared/utils.py
+++ b/suzieq/shared/utils.py
@@ -921,7 +921,7 @@ def print_version():
     print(SUZIEQ_VERSION)
 
 
-def depracated_table_function_warning(dep_table: str, dep_command: str,
+def deprecated_table_function_warning(dep_table: str, dep_command: str,
                                       table: str = None,
                                       command: str = None) -> str:
     """Return the string of the warning for a deprecated function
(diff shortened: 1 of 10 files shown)

### CODE REVIEW

File: suzieq/shared/utils.py [hunk 1]
```diff
@@ -906,3 +906,46 @@ def convert_asndot_to_asn(asn: str) -> int:
 def print_version():
     '''Print the suzieq version and return'''
     print(SUZIEQ_VERSION)
+
+
+def depracated_table_function_warning(dep_table: str, dep_command: str,
```
claudiolor, line 911:
```suggestion
def deprecated_table_function_warning(dep_table: str, dep_command: str,
```

LucaNicosia: thx

claudious96 (approved):

PR #736, opened 2022-06-01
Support pacing commands/logins

ddutt:
When we introduced limiting the max number of outstanding commands to a
AAA server, we didn't pace the commands. As a result, in a network with
fast device responses, we could still overwhelm the AAA server.
Furthermore, some networks only contact the AAA server for
authentication (at login time) while others also contact it for
authorization (per command). Finally, even with the best of
implementations, the AAA server can sometimes fail to
authenticate/authorize even though the credentials are valid. To handle
all these different cases, this patch introduces the following behavior
and configuration knobs:

  * Instead of treating the max-cmd-pipeline parameter as merely
    controlling the max outstanding requests, we instead use this as the
    rate/sec. Thus, a value of 10 is treated as 10 requests/sec which
    then gets converted into a 100ms sleep between requests, thus evenly
    pacing out the 10 requests across the second.
  * We add a per-cmd-auth parameter to the devices section of the
    inventory and use that to control whether we pace commands or only
    logins.
  * We introduce another parameter, retries-on-auth-fail, in the devices
    section of the inventory, and use this to control how many times we
    retry if an authentication fails. This value is reset on successful
    authentication. By default, we don't retry.

## Related Issue

Fixes #

## Description

## Type of change

- Breaking change (fix or feature that would cause existing functionality to not work as expected)
- Documentation update

## New Behavior

...

...

...

...

...

## Comments

Include additional comments about the this pull request

## Double Check

- [ ] I have read the comments and followed the [CONTRIBUTING.md](https://github.com/netenglabs/suzieq/blob/master/CONTRIBUTING.md).
- [ ] I have explained my PR according to the information in the comments or in a linked issue.
- [ ] My PR source branch is created from the `develop` branch.
- [ ] My PR targets the `develop` branch.
- [ ] All my commits have `--signoff` applied

### DIFF
--- a/suzieq/poller/controller/source/base_source.py
+++ b/suzieq/poller/controller/source/base_source.py
@@ -49,6 +49,7 @@ class Source(ControllerPlugin):
             'jump_host_key_file',
             'ignore_known_hosts',
             'slow_host',
+            'per_cmd_auth',
         ]
 
         self._load(input_data)
@@ -204,6 +205,8 @@ class Source(ControllerPlugin):
         transport = None
         ignore_known_hosts = False
         slow_host = False
+        per_cmd_auth = True
+        retries_on_auth_fail = 0
         port = None
         devtype = None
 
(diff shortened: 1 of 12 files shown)


claudiolor (approved):

claudiolor:
I fixed the tests in PR #733 , we might first merge that PR and then update the tests according to these modifications

PR #738, opened 2022-06-01
Support ports in jumphost format (Fix #737)

### LINKED ISSUE #737: [Bug]:  Cannot specify port in Jumphost
### Suzieq version

0.17.2

### Install Type

hand deployed python

### Python version

3.8

### Impacted component

sq-poller

### Steps to Reproduce

1. Specify a jumphost in the inventory under devices
2. Place the port in the URI in the format user@host:port
3. Start poller 
4. observe error

### Expected Behavior

Jumphost is used to connect to devices.

### Observed Behavior

ERROR: 
- devices.proximity:
  + jump-host: format username@jumphost required

### Additional Context

While I will not be using this in production, I can see it being used in other's production setups.

ddutt:

## Related Issue

Fixes #737 

## Description

## Type of change

## New Behavior

Accepts port in jumphost specification (URI format with no scheme). 

...

Adding port to the jumphost specification results in invalid format error.

...

...

Adding docs after sending it to review.

...

jumphost specification now supports port specification as well. Thus the supported URI format is: [<username>@<hostname or ip>[:<port>]

...
## Double Check

### DIFF
--- a/suzieq/poller/controller/utils/inventory_models.py
+++ b/suzieq/poller/controller/utils/inventory_models.py
@@ -1,7 +1,7 @@
 # pylint: disable=no-name-in-module
 
-import re
 from typing import Dict, List, Optional
+from urllib3.util import parse_url
 from pydantic import BaseModel, Field, validator
 
 from suzieq.shared.utils import PollerTransport
@@ -36,11 +36,13 @@ class DeviceModel(BaseModel):
 
     @validator('jump_host')
     def jump_host_validator(cls, value):
-        jump_host = value[2:] if value.startswith('//') else value
-        pattern = '([a-zA-Z0-9-.]{1,253})@([a-zA-Z0-9-.]{1,253})'
-        assert re.fullmatch(pattern, jump_host), \
-               'format username@jumphost required'
-        return value
+        '''Validate jump host format'''
+        try:
+            uri = parse_url(value)
+            assert uri.scheme is None, \
+                'format username@jumphost[:port] required'
+        except ValueError:
+            assert False, f'Invalid jumphost format: {value}'
 
     class Config:
         """pydantic configuration


suzieq/poller/controller/utils/inventory_models.py
```diff
@@ -36,11 +36,13 @@ class DeviceModel(BaseModel):
...
     def jump_host_validator(cls, value):
-        jump_host = value[2:] if value.startswith('//') else value
-        pattern = '([a-zA-Z0-9-.]{1,253})@([a-zA-Z0-9-.]{1,253})'
-        assert re.fullmatch(pattern, jump_host), \
-               'format username@jumphost required'
-        return value
+        '''Validate jump host format'''
+        try:
+            uri = parse_url(value)
+            assert uri.scheme is None, \
```
claudious96, line 42: This validation approach is too broad. For example a value like `ssh://user@` passes the validation since `uri.scheme` is `ssh`. We could just fix the regex to optionally accept the port:
```
pattern = '([a-zA-Z0-9-.]{1,253})@([a-zA-Z0-9-.]{1,253})(\:\d+)?'
```
See this [example](https://regex101.com/r/MCnAGc/1)

claudious96 (changes requested):

suzieq/poller/controller/utils/inventory_models.py
```diff
@@ -36,11 +36,13 @@ class DeviceModel(BaseModel):
...
+                'format username@jumphost[:port] required'
+        except ValueError:
+            assert False, f'Invalid jumphost format: {value}'
```
LucaNicosia, line 45: you should return value otherwise the jump_host won't be set

LucaNicosia (changes requested):

suzieq/poller/controller/utils/inventory_models.py
ddutt: It won't pass because I assert the scheme is None.

ddutt:
```
In [2]: uri = parse_url('ssh://128.0.1.1')

In [3]: assert uri.scheme is None
---------------------------------------------------------------------------
AssertionError                            Traceback (most recent call last)
<ipython-input-3-ff16f5e513e8> in <module>
----> 1 assert uri.scheme is None

AssertionError: 
```

suzieq/poller/controller/utils/inventory_models.py
ddutt: Fixed

suzieq/poller/controller/utils/inventory_models.py
ddutt: I did what you suggested. It doesn't matter which one we pick. Doing what you said stayed closest to the original and is the strictest.

ddutt: Actually, I take it back. The pattern provided doesn't support an IPv6 host. I don't want to reopen this thread when someone uses ipv6. I've gone back to using parse_url.

ddutt:
```
➜ python suzieq/poller/sq_poller.py --syntax-check -I /tmp/inv.yml
2022-06-03 21:31:48,970 - suzieq.poller.controller - WARNING - log level WARNING
2022-06-03 21:31:49,051 - suzieq.poller.controller - ERROR - 
- devices.proximity:
  + jump-host: format username@jumphost[:port] required


suzieq on  fix-737 [$!?] is 📦 v0.18.0a5 via 🐍 v3.8.13 (.venv) 
❯ cat /tmp/inv.yml 
sources:
  - name: LD8-PC-BROA
    hosts:
      - url: ssh://10.1.1.6:22 devtype=eos
      - url: ssh://10.1.1.5:22 devtype=eos
      - url: ssh://10.1.1.11:22 devtype=nxos
      - url: ssh://10.1.1.12:22 devtype=nxos
auths:
  - name: network-ro
    username: ####
    password: ####


devices:
  - name: proximity
    port: 2222
    ignore-known-hosts: true
    jump-host: ssh://ser@172.21.20.2:212
    jump-host-key-file: id_ed25519
```

ddutt:
```
➜ python suzieq/poller/sq_poller.py --syntax-check -I /tmp/inv.yml
2022-06-03 21:32:57,029 - suzieq.poller.controller - WARNING - log level WARNING
Inventory syntax check passed
suzieq on  fix-737 [$!?] is 📦 v0.18.0a5 via 🐍 v3.8.13 (.venv) 
➜ cat /tmp/inv.yml
sources:
  - name: LD8-PC-BROA
    hosts:
      - url: ssh://10.1.1.6:22 devtype=eos
      - url: ssh://10.1.1.5:22 devtype=eos
      - url: ssh://10.1.1.11:22 devtype=nxos
      - url: ssh://10.1.1.12:22 devtype=nxos
auths:
  - name: network-ro
    username: ####
    password: ####


devices:
  - name: proximity
    port: 2222
    ignore-known-hosts: true
    jump-host: user@[2000:1::1]:212
```

PR #744, opened 2022-06-06
Parquet: Handle reading data from multiple sqvers correctly

ddutt:
Reads that involved reading from multiple different sqvers versions was
not handled correctly causing such reads to return incorrect data. This
patch addresses that. Essentially, we need to read each sqvers folder
individually and then join all the dataframes together.

### DIFF
--- a/suzieq/db/parquet/parquetdb.py
+++ b/suzieq/db/parquet/parquetdb.py
@@ -88,6 +88,9 @@ class SqParquetDB(SqDB):
 
         folder = self._get_table_directory(table_name, False)
 
+        if not all(x in fields for x in key_fields):
+            raise ValueError('Key fields MUST be included in columns list')
+
         if addnl_filter:
             # This is for special cases that are specific to an object
             query_str = addnl_filter
@@ -108,7 +111,7 @@ class SqParquetDB(SqDB):
 
         # If requesting a specific version of the data, handle that diff too
         sqvers = kwargs.pop('sqvers', None)
-        datasets = []
+        final_df = pd.DataFrame()
         try:
             dirs = Path(folder)
             try:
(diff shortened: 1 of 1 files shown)


suzieq/db/parquet/parquetdb.py
```diff
@@ -123,64 +123,44 @@ def read(self, table_name: str, data_format: str,
...
-                final_df = final_df.set_index(key_fields) \
-                    .query('~index.duplicated(keep="last")')
+            if not final_df.empty:
+                final_df = final_df.sort_values(by=['timestamp'])
+                dupts_keys = key_fields + ['timestamp']
+                final_df = final_df.set_index(dupts_keys) \
+                    .query('~index.duplicated(keep="last")') \
+                    .reset_index()
+                if (not final_df.empty and (view == 'latest') and
+                        all(x in final_df.columns for x in key_fields)):
```
LucaNicosia, line 161: The check if key_fields are in columns should be done above (before line 157) otherwise we may crash. Cheking it here is useless I believe

LucaNicosia (approved):

suzieq/db/parquet/parquetdb.py
```diff
@@ -620,48 +641,43 @@ def _get_filtered_fileset(self, datasets: list, namespace: list) -> ds:
...
-                         if not x.startswith('!') and not x.startswith('~!')]
-            else:
-                newns = namespace
-            ns_filelist = []
-            chklist = dataset.files
-            for ns in newns or []:
+        if not namespace:
+            return dataset
+
+        notlist = [x for x in namespace
```
claudiolor, line 647: Renaming this variable to `excluded_ns` probably allows us to make code more readeable
```suggestion
        excluded_ns = [x for x in namespace
```

suzieq/db/parquet/parquetdb.py
```diff
@@ -620,48 +641,43 @@ def _get_filtered_fileset(self, datasets: list, namespace: list) -> ds:
...
+                   if x.startswith('!') or x.startswith('~!')]
+        if notlist != namespace:
+            newns = [x for x in namespace
```
claudiolor, line 650: Same here, I would rename to `ns_filters`
```suggestion
            ns_filters = [x for x in namespace
```

suzieq/db/parquet/parquetdb.py
```diff
@@ -620,48 +641,43 @@ def _get_filtered_fileset(self, datasets: list, namespace: list) -> ds:
...
-                newns = namespace
-            ns_filelist = []
-            chklist = dataset.files
-            for ns in newns or []:
+        if not namespace:
+            return dataset
+
+        notlist = [x for x in namespace
+                   if x.startswith('!') or x.startswith('~!')]
+        if notlist != namespace:
```
claudiolor, line 649: I think here a comment explaining why we ignore negations is really useful
```suggestion
        if notlist != namespace:
```

suzieq/db/parquet/parquetdb.py
ddutt: If we don't have all the key fields by this point, we're in trouble. key fields should always be present. Maybe I don't need that check

suzieq/db/parquet/parquetdb.py
```diff
@@ -483,6 +463,47 @@ def _get_avail_sqvers(self, table_name: str, coalesced: bool) -> List[str]:
...
+                         start: str, end: str, fields: List[str],
+                         merge_fields: List[str], query_str: str,
+                         **kwargs) -> pd.DataFrame:
+        '''Process provided dataset and return a pandas DF'''
+
+        # Build the filters for predicate pushdown
+        master_schema = dataset.schema
+
+        avail_fields = list(filter(
+            lambda x: x in master_schema.names, fields))
```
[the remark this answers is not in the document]
ddutt: Ack

suzieq/db/parquet/parquetdb.py
ddutt: Ack

suzieq/db/parquet/parquetdb.py
ddutt: Ack

suzieq/db/parquet/parquetdb.py
ddutt: Added

PR #745, opened 2022-06-06
Path: Always update nexthop IPs for recursive routes, not just for EVPN

ddutt:

### DIFF
--- a/suzieq/engines/pandas/path.py
+++ b/suzieq/engines/pandas/path.py
@@ -449,11 +449,15 @@ class PathObj(SqPandasEngine):
         for row in df.itertuples():
             oifs = []
             nhops = []
-            for nhop in row.nexthopIps:
+            for i, nhop in enumerate(row.nexthopIps):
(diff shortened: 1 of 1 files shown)


suzieq/engines/pandas/path.py
```diff
@@ -462,10 +466,8 @@ def _handle_recursive_route(self, df: pd.DataFrame,
                                                             max_recurse-1)
                     oif = rrdf.oifs.tolist()[0]
                     oifs.extend(oif)
-                    if isinstance(oif, np.ndarray):
-                        oif = oif.tolist()[0]
-                    if oif.startswith('_nexthopVrf:'):
-                        # We need to change the nhop too
+                    if ~rrdf.protocol.isin(['direct', 'local', 'connected',
```
claudiolor, line 469: Why we do here bitwise negation on the result of `.all()`, which is a single True or False. Additionally, it would be useful adding some comments giving an hint of what we are doing, in order to make simpler following the logic

PR #751, opened 2022-06-16
Fix exception handling in the poller

claudiolor:
## Description

The poller sometimes crashes without giving any hint in the log file, this was probably caused by two bugs:

- When a task terminated because of a unhandled exception, the worker run() function didn't properly handle the task exception, ignoring it
- When the poller terminated because an exception, the exception wasn´t written in the log file, but only reported in the stdout 

## Type of change

## New Behavior

When a task terminates because of a unhanded exception, the entire poller crashes, in order not to leave the poller in a half-working state and the exceptions causing the poller termination are written in the log file. Additionally, some log messages in the Node class have been fixed.

### DIFF
--- a/suzieq/poller/worker/nodes/node.py
+++ b/suzieq/poller/worker/nodes/node.py
@@ -463,6 +463,15 @@ class Node:
                 # that sets the device type
                 devtype = self.devtype
             except Exception:
+                self.logger.exception(f'{self.address}:{self.port}: Node '
+                                      'discovery failed due to exception')
+                # All the exceptions related to timeouts and authentication
+                # problems are already catched inside. If we get an
+                # exception here, this is unexpected and most likely something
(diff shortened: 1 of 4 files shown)


suzieq/poller/worker/nodes/node.py
```diff
@@ -463,6 +463,10 @@ async def _detect_node_type(self):
                 # that sets the device type
                 devtype = self.devtype
             except Exception:
+                self.logger.exception('Node discovery failed due to exception')
+                # As there was an exception there is not point in retrying
+                # discovery, it is likely a bug
+                self._retry = 0
```
ddutt, line 468: This exception is not Timeout or any such known error, correct? Maybe adding that to the comment would help.

ddutt (approved):

PR #759, opened 2022-07-01
Alcatel-Lucent Enterprise AOS8 integration

jefvantongerloo:

## Description

## Type of change

...

...

## Double Check

- [x ] My PR source branch is created from the `develop` branch.
- [x ] My PR targets the `develop` branch.


suzieq/poller/worker/services/interfaces.py
```diff
@@ -65,6 +65,65 @@ def _textfsm_valid_speed_value(self, entry):
...
+
+            entry_dict = {}
+
+            adminState = interfaces_status.get('admin_status', -1)
+            ifname = interfaces_status.get('chassis_slot_port', -1)
+            interfaceMac = interfaces.get('mac_address', -1)
+            mtu = interfaces.get('long_frame_size', -1)
+            numChanges = interfaces.get('number_of_status_change', -1)
+            reason = interfaces.get('port_down_violation_reason', -1)
+            speed = interfaces_status.get('detected_speed', -1)
```
claudious96, line 89: Instead of using hardcoded -1, use the `_textfsm_valid_speed_value` method to automatically handle the value and default value in case of missing speed. Just be sure to:
- have the key for `type` in the dictionary passed to the func
- store the speed value in the dictionary as 'speed' instead of 'detected_speed'  and to return an empty string as default value.

suzieq/poller/worker/services/macs.py
```diff
@@ -52,6 +52,40 @@ def _add_mackey_protocol(self, entry):
...
+
+        for entry in processed_data:
+
+            entry_dict = {}
+
+            protocol = entry.get('domain', -1)
+            vlan = entry.get('id', -1)
+            macaddr = entry.get('mac_address', -1)
+            #Where does this parameter go?
+            entry.get('type', -1)
```
claudious96, line 67: I believe this is a leftover, we don't have a `type` value field in the macs table. I think there was a discussion on slack about placing this piece information in the 'flags` field but I'm not sure whether that was cleared out.

suzieq/poller/worker/services/macs.py
```diff
@@ -52,6 +52,40 @@ def _add_mackey_protocol(self, entry):
...
+            else:
+                state = 'suspended'
+            
+            entry_dict = {
+                            'protocol': protocol,
+                            'vlan': vlan,
+                            'macaddr': macaddr,
+                            'flags': flags,
+                            'oif': oif,
+                            'type': type
```
claudious96, line 82: Same as before, we don't have a `type` field in the macs table.

suzieq/poller/worker/services/routes.py
```diff
@@ -36,6 +36,22 @@ def _common_data_cleaner(self, processed_data, _):
...
+
+        entry_new = []
+
+        for entry in processed_data:
+
+            entry_dict = {
+                'prefix': entry['dest_addr'],
+                'nexthopIps': [entry['gateway_addr']],
+                'protocol': entry['protocol'],
+            }
```
claudious96, line 49: Would it be possible to retrieve also fields like: `oifs`, `source`, `vrf`, `action`?

suzieq/poller/worker/services/interfaces.py
```diff
@@ -65,6 +65,65 @@ def _textfsm_valid_speed_value(self, entry):
...
+                'interfaceMac': interfaceMac,
+                'mtu': mtu,
+                'numChanges': numChanges,
+                'reason': reason,
+                'speed': speed,
+                'state': state,
+                'type': type,
+                'vlan': vlan,
+                'vlanList' : vlanList
+            }
```
claudious96, line 121: Would it be possible to retrieve also `ipAddressList` and `master` (vrf) fields?

suzieq/poller/worker/services/arpnd.py
```diff
@@ -10,6 +10,16 @@
...
 
+    def _clean_aos_data(self, processed_data, _):
+
+        entry, = processed_data
+
+        entry['ipAddress'] = entry.get('ip_address', '-1')
+        entry['macaddr'] = entry.get('hardware_addr', '-1')
+        entry['oif'] = entry.get('port', '-1') # or 'port' for physical interface
+
+        return processed_data
```
claudious96, line 21: Do you think it could be possible to retrieve also the `state` field (reachable, permanent .. )?

claudious96 (changes requested):
First of all thanks for the work!

I left some minor comments about styling or linting-related. You can see the reason in the github actions for flake8 and pylint, or just run locally `pylint suzieq` and `flake8 suzieq`.

Others are related to default values in the cleanup functions in services and about some missing fields for tables entries: arpnd, interfaces, routes.

suzieq/poller/worker/nodes/node.py
```diff
@@ -1277,6 +1285,45 @@ def _extract_nos_version(self, data) -> None:
...
+class AosNode(Node):
+    '''Alcatel AOS Node-specific implementation'''
+
+    async def _parse_init_dev_data(self, output, cb_token) -> None:
+        
+        if output[0]['status'] == 0:
+            data = output[0]['data']
+
+            # Extract hostname
+            hname = re.search('\s+Name:\s+(\S+),', data)
```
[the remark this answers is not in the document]
jefvantongerloo: Added raw string literal for all regex functions.

suzieq/poller/worker/services/macs.py
jefvantongerloo: Adjusted to use the function, but gives the same results as before.

suzieq/poller/worker/services/interfaces.py
jefvantongerloo: Adjusted to use the function, but gives the same results as before.

jefvantongerloo:
Forgot to signoff latest commits and rebased with --signoff parameter.

suzieq/poller/worker/services/macs.py
jefvantongerloo: Removed `type` field.

suzieq/poller/worker/services/macs.py
jefvantongerloo: Removed `type` field.

suzieq/poller/worker/nodes/node.py
```diff
@@ -494,7 +499,8 @@ async def _get_device_type_hostname(self):
         # setup time. show version works on most networking boxes and
         # hostnamectl on Linux systems. That's all we support today.
         await self._exec_cmd(self._parse_device_type_hostname,
-                             ["show version",
+                             ["show system",
+                              "show version",
```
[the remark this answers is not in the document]
jefvantongerloo: This really should be fixed with a durable solution.

There is a problem with the ´only_one´ parameters on the cmd execution in function ´_get_device_type_hostname´ for the device discovery. If an unknown command is sent to the device, status 0 and an error message in standard output get returned. So when the correct command for that type of device is not used first, the device gets discovered as unknown.

The program can also never enter the elif function because only one command from the list is sent if the return status is 0 (even with an error return):

```
        elif (len(output) > 1) and (output[1]["status"] == 0):
            devtype = 'unsupported'
            data = output[1]["data"]
            if data:
                if "Cumulus Linux" in data:
                    devtype = "cumulus"
                else:
                    devtype = "linux"

                version_str = data
                # Hostname is the last line of the output
                if len(data.strip()) > 0:
                    hostname = data.splitlines()[-1].strip()
```

Sample terminal output on unknown cli command (with return status code 0):

```
net-swi-001> show version
                     ^
ERROR: Invalid entry: "version"
```

suzieq/poller/worker/services/arpnd.py
jefvantongerloo: Added ´state´ column from ´type´ field:
STATIC = permanent
DYNAMIC = reachable

suzieq/poller/worker/nodes/node.py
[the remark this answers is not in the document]
claudious96: > This really should be fixed with a durable solution.

I agree. What you propose, checking for a command failure using both the return code AND a regex for the output, seems the best way to go.

For this first phase I'd go for supporting Alcatel-Lucent only supplying `devtype` as `aos` in the inventory, we can implement the new way of determining a failed command separately and then come back to having the autodiscovery.

> Sample terminal output on unknown cli command (with return status code 0):
> 
> net-swi-001> show version
>                      ^
> ERROR: Invalid entry: "version"

Does this appear in the stdout or stderr?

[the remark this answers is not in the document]
jefvantongerloo: Terminal output from cli command, so stdout.

suzieq/poller/worker/services/routes.py
jefvantongerloo: `oifs` could be added, but another more complex command has to be parsed: `show ip router database`

jefvantongerloo:
- Updated and tested the textfsm templates in latest aos releases (8.9.73.R01 & 8.8.56.R02).
- Added ip interfaces en aggregate interfaces
- Added ip and vlan members on route and interface services
- Added oif on router service

PR #761, opened 2022-07-08
Network find vlan filter fix

LucaNicosia:
## Description

The vlan filter in the `network find` function wasn't working properly. Since the vlan field in the dataframe was converted to a string, the query `vlan == {vlan}` will try to compare an integer with a string, resulting into never matching.

Moreover, a check for vlan to be an integer is now added.

## Type of change

### DIFF
--- a/suzieq/engines/pandas/network.py
+++ b/suzieq/engines/pandas/network.py
@@ -79,6 +79,12 @@ class NetworkObj(SqPandasEngine):
         columns = kwargs.get('columns', ['default'])
         cols = self.schema.get_display_fields(columns)
 
+        if vlan:
+            try:
+                vlan = int(vlan)
+            except ValueError:
+                return pd.DataFrame({'error': ['vlan must be an integer']})
+
         # Convert Cisco-style MAC address to standard MAC addr format,
         # and lowercase all letters of the alphabet
         addr = convert_macaddr_format_to_colon(addr)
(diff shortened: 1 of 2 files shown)


suzieq/engines/pandas/network.py
```diff
@@ -179,6 +185,15 @@ def _find_addr_arp(self, addr: str, **kwargs) -> pd.DataFrame:
...
             if not ifdf.empty:
+                vrf = ifdf.master.unique().tolist()[0] or 'default'
+                vlan_list = ifdf.vlan.astype(str).unique().tolist()
+                try:
+                    vlan = vlan_list[0]
+                    vlan = int(vlan)
+                except ValueError:
+                    # ignore if the conversion of the vlan to integer has
+                    # failed
+                    pass
```
claudiolor, line 196: These lines are not clear to me. Why we need to perform cast to string with `.astype(str)` and then we cast back to int in line 192?

LucaNicosia: I actually do not have a response to this. I simply moved the code without modifying it. I guessed it was a problem of NaN values in the unique function, but it seems not.
Maybe @ddutt can give us a better answer

ddutt: This looks like its been added by you @LucaNicosia ?

ddutt: In places where VLAN was not valid such as a routed link, I was putting "-" and so I made the vlan as string. Luca did half a fix :)

ddutt:
Wrong fix. Fixing via separate PR

PR #764, opened 2022-07-25
Mackey update

ddutt:
When we changed OIFs with a name like sup-eth1(R) to sup-eth1 with a flag set to router, we didn't account for the fact that the mackey needed changing to avoid duplicating mac address entries that shouldn't be duplicates. This patch addresses this issue.

We also fix tests to cover the case now. Additionally tests that should've been updated due to the removal of certain unused services such as topmem and ifcounters have been updated. 

A final patch to this PR is the correct handling of a service specification in update_tests script.


suzieq/poller/worker/services/macs.py
```diff
@@ -45,9 +45,12 @@ def _add_mackey_protocol(self, entry):
...
             else:
                 if entry.get('vlan', 0):
-                    entry['mackey'] = entry['vlan']
+                    if entry['flags'] in ['static', 'permanent', 'router']:
+                        entry['mackey'] = f'{entry["vlan"]}-{entry["oif"]}'
+                    else:
+                        entry['mackey'] = entry['vlan']
                 else:
-                    if entry['flags'] in ['static', 'permanent']:
+                    if entry['flags'] in ['static', 'permanent', 'router']:
```
LucaNicosia, line 53: What about older records? Don't we need some kind of "migration function" to update the mackeys?

ddutt: Excellent question. The bug was introduced in 0.18.0 and only affects NXOS. Its not a critical bug in the sense that none of the critical functionality will be broken as a consequence of this. I could provide a migration function, I suppose.

claudious96 (approved):

claudiolor (approved):

PR #766, opened 2022-07-27
Nwfind fix

ddutt:
Network find had a bunch of fixes:

* Correct handling when both start and end times are provided or view=all
* Correct tagging of an address as bridged/routed
* Correct handling of an address that is an interface address on a device we're polling.

### DIFF
--- a/suzieq/engines/pandas/network.py
+++ b/suzieq/engines/pandas/network.py
@@ -87,14 +87,13 @@ class NetworkObj(SqPandasEngine):
         if addr_df.empty:
             # Is this a locally attached interface IP to a device we're polling
             df = self._get_table_sqobj('address') \
-                .get(vrf=vrf, address=[addr], columns=['*'],
+                .get(vrf=vrf, address=[addr],
+                     columns=['namespace', 'hostname', 'ifname', 'vlan', 'vrf',
+                              'ipAddress', 'macaddr'],
                      **kwargs)
             if df.empty:
                 return addr_df
 
-            df = df.drop(columns=['ipAddressList', 'ip6AddressList',
-                                  'state', 'type'],
-                         errors='ignore')
             df['ipAddress'] = df.ipAddress.apply(
                 lambda x: ', '.join([y.split('/')[0] for y in x]))
 
@@ -102,6 +101,7 @@ class NetworkObj(SqPandasEngine):
             df['type'] = 'interface'
             df['l2miss'] = False
 
+            cols = [x for x in cols if x in df.columns]
             return df[cols]
 
         addr_df = self._find_first_hop_attach(addr_df)
@@ -213,7 +213,7 @@ class NetworkObj(SqPandasEngine):
                 if not ifdf.ipAddressList.apply(
                         lambda subnets, netaddr:
                         False if not subnets.any() else
-                        any(addr in ip_network(subnet, strict=False)
+                        any(netaddr in ip_network(subnet, strict=False)
                             for subnet in subnets),
                         args=(netaddr,)).any():
                     # Routed interface
--- a/suzieq/engines/pandas/network.py
+++ b/suzieq/engines/pandas/network.py
@@ -1,4 +1,5 @@
 from typing import List
+from ipaddress import ip_address, ip_network
 
 import pandas as pd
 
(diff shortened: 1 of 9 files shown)


suzieq/engines/pandas/network.py
```diff
@@ -179,7 +206,15 @@ def _find_addr_arp(self, addr: str, **kwargs) -> pd.DataFrame:
...
-                if 'vlan' not in ifdf.type.unique():
+                netaddr = ip_address(row.ipAddress)
+                # The code below checks if the provided address belongs to
+                # the interface subnet. If it does, its a bridged address
+                # else its a routed address i.e. you use routing to reach it.
+                if not ifdf.ipAddressList.apply(
+                        lambda x, addr:
+                        False if not x.any() else
+                        any(addr in ip_network(y, strict=False) for y in x),
+                        args=(netaddr,)).any():
```
claudiolor, line 217: This piece of code is not easy to follow, I would rename some of the variables:
```suggestion
                if not ifdf.ipAddressList.apply(
                        lambda addr_list, netaddr:
                        False if not addr_list.any() else
                        any(netaddr in ip_network(a, strict=False) for a in addr_list),
                        args=(netaddr,)).any():
```

ddutt: Fixed with diff var names

suzieq/engines/pandas/network.py
```diff
@@ -75,14 +76,34 @@ def _find_address(self, addr: str, **kwargs) -> pd.DataFrame:
...
+                                  'state', 'type'],
+                         errors='ignore')
+            df['ipAddress'] = df.ipAddress.apply(
+                lambda x: ', '.join([y.split('/')[0] for y in x]))
+
+            df['bondMembers'] = ''
+            df['type'] = 'interface'
+            df['l2miss'] = False
+
+            return df[cols]
```
LucaNicosia, line 105: I think if one of the network columns is not in address, this line will fail

suzieq/engines/pandas/network.py
```diff
@@ -179,7 +206,16 @@ def _find_addr_arp(self, addr: str, **kwargs) -> pd.DataFrame:
...
             if not ifdf.empty:
-                if 'vlan' not in ifdf.type.unique():
+                netaddr = ip_address(row.ipAddress)
+                # The code below checks if the provided address belongs to
+                # the interface subnet. If it does, its a bridged address
+                # else its a routed address i.e. you use routing to reach it.
+                if not ifdf.ipAddressList.apply(
+                        lambda subnets, netaddr:
+                        False if not subnets.any() else
+                        any(addr in ip_network(subnet, strict=False)
```
LucaNicosia, line 216: Both `ip_address(row.ipAddress)` and `ip_network(subnet, strict=False)` expect a `_RawIpAddress`. `ip_address` seems to work also with a string, but `ip_network` raises this exception:
```shell
luca> network find address='10.0.0.11'                                                                                                                                                                                                                                                                Error: 1
                                             error
0  ERROR: 'str' object has no attribute '_version'
```

LucaNicosia (changes requested):
Some network tests are failing. The reason is the `ip_network` bug

suzieq/engines/pandas/network.py
```diff
@@ -75,14 +76,34 @@ def _find_address(self, addr: str, **kwargs) -> pd.DataFrame:
...
+            # Is this a locally attached interface IP to a device we're polling
+            df = self._get_table_sqobj('address') \
+                .get(vrf=vrf, address=[addr], columns=['*'],
+                     **kwargs)
+            if df.empty:
+                return addr_df
+
+            df = df.drop(columns=['ipAddressList', 'ip6AddressList',
+                                  'state', 'type'],
+                         errors='ignore')
```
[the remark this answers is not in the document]
ddutt: Done

suzieq/engines/pandas/network.py
ddutt: Fixed. Typo when I changed var names.

suzieq/engines/pandas/network.py
ddutt: Fixed

PR #777, opened 2022-08-08
Netbox multi tag

LucaNicosia:
## Description

This PR adds the possibility to use multiple tags on netbox source.

Now netbox tags should be specified as a list. For backward compatibility, the string format is only marked as deprecated but still accepted.

## Type of change

## Double Check

### DIFF
--- a/suzieq/poller/controller/source/netbox.py
+++ b/suzieq/poller/controller/source/netbox.py
@@ -11,7 +11,7 @@ Classes:
 
 import asyncio
 import logging
-from typing import Dict, List, Optional, Tuple, Union
+from typing import Any, Dict, List, Optional, Tuple, Union
 from urllib.parse import urlparse
 
 from pydantic import BaseModel, validator, Field
@@ -44,7 +44,7 @@ class NetboxServerModel(BaseModel):
 class NetboxSourceModel(SourceModel):
     """Netbox source validation model
     """
-    tag: Optional[str] = Field(default='suzieq')
+    tag: Optional[Any] = Field(default=['suzieq'])
     period: Optional[int] = Field(default=3600)
     token: str
     ssl_verify: Optional[bool] = Field(alias='ssl-verify')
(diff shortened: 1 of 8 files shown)


suzieq/poller/controller/source/netbox.py
```diff
@@ -160,16 +191,37 @@ async def get_inventory_list(self) -> List:
...
+                while next_url:
+                    logger.debug(f"Netbox: Retrieving url '{next_url}'")
+                    cur_devices, next_url = await self._get_devices(next_url)
+                    url_devices.extend(cur_devices)
+                for device in url_devices:
+                    # remove duplicates
+                    dev_id = device.get('id')
+                    if dev_id is not None and dev_id not in ids:
+                        ids.add(dev_id)
+                        devices.append(device)
```
claudiolor, line 224: We might simplify this piece of code just building a dictionary and returning the values of the dictionary:
```suggestion
               # Remove the duplicates from the list of received devices
              devices = list({device['id']: device for device in url_devices
                              if device.get('id') is not None}.values())
```

suzieq/poller/controller/source/netbox.py
```diff
@@ -151,6 +164,24 @@ def _token_auth_header(self) -> Dict:
...
+
+        Returns:
+            List[str]: list of urls
+        """
+        urls = []
+        url_address = f'{self._server.protocol}://{self._server.host}:'\
+            f'{self._server.port}/api/dcim/devices/?'
+        for tag in self._data.tag:
+            query = url_address
+            for (i, t) in enumerate(tag.split(", ")):
```
claudiolor, line 178: This works only when you add a space after the comma: `one, two`. A solution might be splitting by comma and then stripping the remaining spaces, if any:
```suggestion
            match_tags = [t.strip() for t in tag.split(',')]
            for i, t in enumerate(match_tags):
``

File: suzieq/poller/controller/source/netbox.py [hunk 3]
```diff
@@ -151,6 +164,24 @@ def _token_auth_header(self) -> Dict:
...
+    def _get_url_list(self) -> List[str]:
+        """Return the list of requests to execute
+
+        Returns:
+            List[str]: list of urls
+        """
+        urls = []
+        url_address = f'{self._server.protocol}://{self._server.host}:'\
+            f'{self._server.port}/api/dcim/devices/?'
+        for tag in self._data.tag:
```
claudiolor, line 176:
```suggestion
        for tags in self._data.tag:
```
Maybe we should use plural var name as each row might have more than one tag

suzieq/poller/controller/source/netbox.py
LucaNicosia: This solution is not possible as is because devices must contain unique devices for each different tag. With the solution you proposed, `devices` will contain the result of only the last url. Anyway, I'll simplify the code with a solution similar to your suggestion. Thanks

claudious96 (approved):

ddutt (approved):

PR #779, opened 2022-08-09
Poller fixes 19

ddutt:
This patch addresses several fixes:

* Removes the IPv6 loopback address from being saved for Linux-y NOS (Cumulus, SONiC)
* Treats the .0 interface in Junos to be of the same type as the parent interface (lo0.0 is the same type as lo0, but not lo0.1)
* Not printing "Ignoring service" for blacklisted services
* mac show displaying the correct information when user requests viewing moveCount logical column
* More improvements to network find
* Display data sorted by timestamp when user specifies view=all or provides both start and end times

I'll update the tests (moveCount tests are failing right now) and add new tests for network find once the review is done.

### DIFF
--- a/suzieq/cli/sqcmds/command.py
+++ b/suzieq/cli/sqcmds/command.py
@@ -274,6 +274,7 @@ class SqCommand(SqPlugin):
         else:
             print(df)
 
+    # pylint: disable=too-many-statements
     def _gen_output(self, df: pd.DataFrame, json_orient: str = "records",
                     dont_strip_cols: bool = False, sort: bool = True):
 
@@ -332,8 +333,13 @@ class SqCommand(SqPlugin):
                     if is_error:
                         print(df[cols])
                     else:
-                        sort_fields = [x for x in self.sqobj.sort_fields
-                                       if x in df.columns and x in cols]
+                        if (((self.start_time and self.end_time) or
+                             self.view == "all") and
+                                'timestamp' in df.columns):
+                            sort_fields = ['timestamp']
+                        else:
+                            sort_fields = [x for x in self.sqobj.sort_fields
+                                           if x in df.columns and x in cols]
                         if sort_fields:
                             self._pager_print(
                                 df[cols].sort_values(by=sort_fields,
--- a/suzieq/engines/pandas/network.py
+++ b/suzieq/engines/pandas/network.py
@@ -1,6 +1,5 @@
 from typing import List
 from ipaddress import ip_address, ip_network
-from datetime import timedelta
 
 import pandas as pd
 
@@ -165,6 +164,7 @@ class NetworkObj(SqPandasEngine):
 
         return [ifname]
 
+    # pylint: disable=too-many-statements
     def _find_addr_arp(self, addr: str, **kwargs) -> pd.DataFrame:
         """Find the origin or the first hop network device that owns this addr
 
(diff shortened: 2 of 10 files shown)


suzieq/engines/pandas/network.py
```diff
@@ -204,20 +209,33 @@ def _find_addr_arp(self, addr: str, **kwargs) -> pd.DataFrame:
...
-                oif = arp_row.oif
+                # Ignore VRR interfaces in Cumulus, we got them via the
+                # normal interface entry (without -v0)
+                continue
+            oif = arp_row.oif
 
             active = getattr(arp_row, 'active', True)
+            if (arp_row.remote and
+                    arp_row.hostname not in active_local_hosts):
+                continue
```
LucaNicosia, line 231: Can you add a comment here? Now we know what this 'if' does, but in the future we'll forget.

LucaNicosia (approved):

suzieq/engines/pandas/network.py
```diff
@@ -204,20 +209,33 @@ def _find_addr_arp(self, addr: str, **kwargs) -> pd.DataFrame:
...
+
+            # If an entry goes from local to being deleted, remove it
+            if not active and arp_row in active_local_hosts:
```
claudious96, line 234:
```suggestion
            if not active and arp_row.hostname in active_local_hosts:
```

claudious96 (changes requested):

suzieq/cli/sqcmds/command.py
```diff
@@ -332,8 +332,13 @@ def _gen_output(self, df: pd.DataFrame, json_orient: str = "records",
...
                     else:
-                        sort_fields = [x for x in self.sqobj.sort_fields
-                                       if x in df.columns and x in cols]
+                        if (((self.start_time and self.end_time) or
+                             self.view == "all") and
+                                'timestamp' in df.columns):
+                            sort_fields = ['timestamp']
+                        else:
+                            sort_fields = [x for x in self.sqobj.sort_fields
+                                           if x in df.columns and x in cols]
```
claudious96, line 341: Just a general comment, probably unrelated specifically to this PR. Shouldn't we do the ordering in the engine instead of the CLI? As it is now we may have different results for the same commands in the CLI vs the rest server since the final ordering happens here.

suzieq/engines/pandas/network.py
```diff
@@ -200,23 +205,37 @@ def _find_addr_arp(self, addr: str, **kwargs) -> pd.DataFrame:
...
+                active_local_hosts.remove(arp_row.hostname)
+
+            if active and not arp_row.remote:
+                active_local_hosts.add(arp_row.hostname)
 
             # We ignore the start time because the interface may not have
```
claudiolor, line 240: This comment seems a leftover. Probably it should be moved up to line 205

suzieq/cli/sqcmds/command.py
ddutt: In general, I leave ordering to be a front-end thingy since with a lot of data, ordering takes time.

suzieq/engines/pandas/network.py
```diff
@@ -182,13 +181,19 @@ def _find_addr_arp(self, addr: str, **kwargs) -> pd.DataFrame:
...
             return pd.DataFrame(result)
 
-        arpdf = arpdf.query('not remote').reset_index(drop=True)
-        if arpdf.empty:
-            return pd.DataFrame(result)
+        if not arpdf.empty:
+            key_cols = ['namespace', 'hostname', 'ipAddress']
+            arpdf = arpdf.sort_values(by=key_cols + ['timestamp'])
+            arpdf['next_ts'] = arpdf.groupby(by=key_cols)['timestamp'] \
+                                    .shift(-1)
```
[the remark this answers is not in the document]
ddutt: Ack, adding.

suzieq/engines/pandas/network.py
ddutt: Ack

suzieq/engines/pandas/network.py
ddutt: OK

suzieq/engines/pandas/network.py
ddutt: Ugh, horrible bug. Nice catch

claudious96 (approved):

PR #782, opened 2022-08-12
Bump pyarrow up to version 8.0

claudiolor:
## Description

This PR bumps the version of pyarrow up to 8.0. Additionally we solved some deprecation:
- Bump the version of Parquet format from 2.0 to 2.4
- Replace the `partition_filename_cb arg` with `basename_template`

Additionally, we solved a bug causing the migration function to fail when there was no data in the directory.

## Type of change

- Dependency upgrade
- Bug fix


suzieq/db/parquet/parquetdb.py
```diff
@@ -609,7 +613,7 @@ def _get_cp_dataset(self, table_name: str, need_sqvers: bool,
         if filelist:
             return ds.dataset(filelist, format='parquet', partitioning='hive')
         else:
-            return []
+            return None
```
ddutt, line 614: Why None instead of empty list?

suzieq/db/parquet/parquetdb.py
```diff
@@ -224,13 +224,13 @@ def write(self, table_name: str, data_format: str,
             if filename_cb:
                 pq.write_to_dataset(table, root_path=folder,
                                     partition_cols=partition_cols,
-                                    version="2.0", compression="ZSTD",
+                                    version="2.4", compression="ZSTD",
```
ddutt: Can I be Lorina and say why not use a #define constant to hold the parquet version :)

suzieq/db/parquet/pq_coalesce.py
```diff
@@ -40,7 +41,11 @@ def pq_file_name(self, *args):  # pylint: disable=unused-argument
 
         """
         # Using timestamp rather than date/time string to simplify reads
-        return f'{self.prefix}{self.block_start}-{self.block_end}.parquet'
+        name_suffix = f'{self.block_start}-{self.block_end}.parquet'
+
+        # {i} is replaced by pyarrow with a sequential number. It must be
+        # provided when a name for the output parquet file is provided
+        return f'{self.prefix}' + '{i}-' + name_suffix
```
ddutt, line 48: Why not use a single f-string instead of using +?

ddutt (changes requested):

suzieq/db/parquet/parquetdb.py
claudiolor: Because we return a dataset as result not a list, why should we return an empty list if normally don't return a list?

suzieq/db/parquet/pq_coalesce.py
claudiolor: I can't use fstring in this case, as otherwise `i` is interpreted as a variable

PR #783, opened 2022-08-15
More misc fixes 19

ddutt:
This patchset fixes the following issues:

- VLAN filter with interfaces wasn't working correctly if you passed any operators such as !, <, >= etc.
- Any numeric filter that took an operator had to be given without a space between the operator and the number (>1500, not > 1500). This caused weird errors being reported which was confusing and not indicative of the problem. Instead of detecting this to throw a good error, this patch just fixes it
- The fix above had side-effects percolating into other areas of code, but none were breaking changes
- WHen the user specifies both </<= and >/>=, it typically makes sense only as an AND, i.e. the user is looking for a number between the two. This is now handled correctly
Tests have been updated to handle the first two cases. Need to add tests for the last bullet

### DIFF
--- a/suzieq/config/schema/topology.avsc
+++ b/suzieq/config/schema/topology.avsc
@@ -57,13 +57,13 @@
         },
         {
             "name": "asn",
-            "type": "string",
+            "type": "long",
             "display": 7,
             "description": "BGP ASN for BGP peering"
         },
         {
             "name": "peerAsn",
-            "type": "string",
+            "type": "long",
             "display": 8,
             "description": "BGP peerASN in BGP peering"
         },
--- a/suzieq/db/parquet/parquetdb.py
+++ b/suzieq/db/parquet/parquetdb.py
@@ -687,15 +687,15 @@ class SqParquetDB(SqDB):
 
         # Check if we have logical operator (<, >, = etc.)
         if filter_str.startswith('<='):
-            return (ds.field(keyfld) <= int(filter_str[2:]))
+            return (ds.field(keyfld) <= int(filter_str[2:].strip()))
         elif filter_str.startswith('>='):
-            return (ds.field(keyfld) >= int(filter_str[2:]))
+            return (ds.field(keyfld) >= int(filter_str[2:].strip()))
         elif filter_str.startswith('<'):
-            return (ds.field(keyfld) < int(filter_str[1:]))
+            return (ds.field(keyfld) < int(filter_str[1:].strip()))
         elif filter_str.startswith('>'):
-            return (ds.field(keyfld) > int(filter_str[1:]))
+            return (ds.field(keyfld) > int(filter_str[1:].strip()))
         else:
-            return (ds.field(keyfld) == int(filter_str))
+            return (ds.field(keyfld) == int(filter_str.strip()))
 
     # pylint: disable=too-many-statements
     def build_ds_filters(self, start_tm: float, end_tm: float,
@@ -731,7 +731,15 @@ class SqParquetDB(SqDB):
             if isinstance(v, list):
                 infld = []
                 notinfld = []
-                or_filters = None
+                kw_filters = None
+                use_and = False
+                # If user specifies both </<= and >/>=, we treat
+                # it as an and i.e. the user is looking for a val
+                # between the provided < and > numbers.
+                if (any(x.startswith('>') for x in v) and
+                        any(x.startswith('<') for x in v)):
+                    use_and = True
+
                 for e in v:
                     if isinstance(e, str) and e.startswith("!"):
                         if ftype == 'int64':
@@ -740,11 +748,15 @@ class SqParquetDB(SqDB):
                             notinfld.append(e[1:])
                     else:
                         if ftype == 'int64':
-                            if or_filters is not None:
-                                or_filters = or_filters | \
-                                    self._cons_int_filter(k, e)
+                            if kw_filters is not None:
+                                if use_and:
+                                    kw_filters = kw_filters & \
+                                        self._cons_int_filter(k, e)
+                                else:
+                                    kw_filters = kw_filters | \
+                                        self._cons_int_filter(k, e)
                             else:
-                                or_filters = self._cons_int_filter(k, e)
+                                kw_filters = self._cons_int_filter(k, e)
                         else:
                             infld.append(e)
                 if infld and notinfld:
@@ -755,8 +767,8 @@ class SqParquetDB(SqDB):
                 elif notinfld:
                     filters = filters & (~ds.field(k).isin(notinfld))
 
-                if or_filters is not None:
-                    filters = filters & (or_filters)
+                if kw_filters is not None:
+                    filters = filters & (kw_filters)
             else:
                 if isinstance(v, str) and v.startswith("!"):
                     if ftype == 'int64':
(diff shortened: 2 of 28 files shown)


suzieq/sqobjects/basicobj.py
```diff
@@ -117,10 +119,15 @@ def _check_input_for_valid_args(self, good_arg_list, **kwargs,):
...
-            return
-
+        fields = self.schema.fields
         for arg, val in kwargs.items():
+
+            # if we have integer fields, make sure we catch things like
+            # user specifying "> 25" and treating them as a single val
+            # instead of 2 separate args due to the split on space
+            if arg not in fields:
+                continue
```
claudious96, line 129: The comment seems unrelated to what happens here.

ddutt: Ack, removing.

suzieq/config/schema/topology.avsc
```diff
@@ -57,7 +57,7 @@
         },
         {
             "name": "asn",
-            "type": "string",
+            "type": "long",
```
LucaNicosia, line 60: There is also the peerAsn field which is a string instead of a long. Should we update that as well?

suzieq/engines/pandas/interfaces.py
```diff
@@ -89,6 +87,69 @@ def get(self, **kwargs):
...
+            elif vlan.startswith(('<', '>', '!')):
+                op = vlan[0]
+                vlan = vlan[1:].strip()
+            else:
+                op = '=='
+
+            vlan = int(vlan)
+            if op == "!":
+                tmpdf = df[df.apply(
+                    lambda x, vlan: all(opdict[op](y, vlan)
```
claudiolor, line 135: Can you replace `x` and `y` with some meaningful names?

suzieq/engines/pandas/interfaces.py
```diff
@@ -89,6 +87,69 @@ def get(self, **kwargs):
...
+                                        for y in x.vlanList) and
+                    opdict[op](x.vlan, vlan), args=(vlan,), axis=1)]
+            else:
+                tmpdf = df[df.apply(
+                    lambda x, vlan: any(opdict[op](y, vlan)
```
claudiolor, line 140: If we directly assign the operation to `op` we can directly do `op(..)`, which I think is more readeable

suzieq/engines/pandas/interfaces.py
```diff
@@ -89,6 +87,69 @@ def get(self, **kwargs):
...
+        # If there are any VLANs with ! mixed without !, remove the
+        # ! ones. vlanList is an OR list, not an AND list, except
+        # if all are !
+        if any(x.startswith('!') for x in vlan_list):
+            cond = 'and'
+        else:
+            cond = 'or'
+
+        resdf_list = []
+        if (any(x.startswith('>') for x in vlan_list) and
```
claudiolor, line 112: Wondering whether this is the right thing to do. 
It is true in the case of a condition like `vlan > 10 vlan < 20` we need an `and`, but it also true that if I want something like `vlan < 10 vlan > 20` I need an `or`. 
Can't we find a way to make things more explicit? 
Maybe we can ask to specify `and` or `or` or we might define an operator for intervals, 
for example id we want an interval we can say `[10, 20]` for `10 <= vlan <= 20`  or `]10, 20]` for `10 < vlan <= 20`.

suzieq/db/parquet/parquetdb.py
```diff
@@ -731,7 +731,15 @@ def build_ds_filters(self, start_tm: float, end_tm: float,
             if isinstance(v, list):
                 infld = []
                 notinfld = []
-                or_filters = None
+                kw_filters = None
+                use_and = False
+                # If user specifies both </<= and >/>=, we treat
+                # it as an and i.e. the user is looking for a val
+                # between the provided < and > numbers.
+                if (any(x.startswith('>') for x in v) and
```
claudiolor, line 739: Same as previous comment

suzieq/engines/pandas/interfaces.py
```diff
@@ -89,6 +87,69 @@ def get(self, **kwargs):
...
+                any(x.startswith('<') for x in vlan_list)):
+            # if the user specifies both >/>= and </<=, then we
+            # treat this as an and
+            cond = 'and'
+
+        opdict = {'<': operator.lt, '>': operator.gt,
```
claudiolor, line 118: We defined this dictionary in many places. Can't we just define a function in utils doing this mapping everywhere? (or I think also storing the dictionary once in a shared location is fine)

suzieq/sqobjects/basicobj.py
```diff
@@ -129,15 +133,30 @@ def _check_input_for_valid_vals(self, good_arg_val_list, **kwargs):
...
+        fields = self.schema.fields
+        for arg, val in kwargs.items():
+            if arg not in fields or not val or not isinstance(val, list):
+                # only if the value can be a split does a "> 100" become
+                # a two element list instead of being a single element.
+                # This is what the code below is fixing
+                continue
+            if (self.schema.field(arg).get('type', '') in ['int', 'long',
+                                                           'float']):
+                user_val = ' '.join(val)
```
claudiolor, line 151: While we are here, can't we check whether the syntax of the conditions is valid?

suzieq/engines/pandas/interfaces.py
```diff
@@ -89,6 +87,69 @@ def get(self, **kwargs):
...
+                    opdict[op](x.vlan, vlan), args=(vlan,), axis=1)]
+            if cond == "or":
+                resdf_list.append(tmpdf.reset_index(drop=True))
+            else:
+                df = tmpdf
+
+        if resdf_list:
+            df = pd.concat(resdf_list)
+
+        return df.query('vlanList.str.len() > 0 or vlan != 0')
```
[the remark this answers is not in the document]
ddutt: vlan = 0 is not valid as a filter. Its a fill-in value when its a routed port or some other kind of port

suzieq/engines/pandas/interfaces.py
ddutt: I thought we had, but I didn't find it. Will look again.

suzieq/engines/pandas/interfaces.py
ddutt: Was trying to avoid using up another variable when I use it only once. Its not that unreadable

suzieq/engines/pandas/interfaces.py
ddutt: lambda x, y is a pretty commonly used pattern, though I agree that meaningful names makes sense. Its just that in this case, I'm not sure it makes the code that much more readable. This is not a complicated piece of code.

suzieq/config/schema/topology.avsc
ddutt: Yes! I missed that. Thanks

suzieq/engines/pandas/interfaces.py
claudiolor: Maybe we can give `r` for row instead of `x` and and `v` for vlan instead of `y`. They are short but at least they give an hint of what they are.

ddutt: Done

suzieq/engines/pandas/interfaces.py
ddutt: Its only in device.py and here. Will see if I can move them into engineobj

suzieq/sqobjects/basicobj.py
ddutt: Separately. Lets open a ticket on that. It maybe more work than I want to take on in this patch. I want to put out 0.19.0

PR #792, opened 2022-09-01
Add query_str to path

LucaNicosia:
## Description

The `query_str` parameter was missing in path

## Type of change

## Double Check

### DIFF
--- a/suzieq/cli/sqcmds/PathCmd.py
+++ b/suzieq/cli/sqcmds/PathCmd.py
@@ -25,6 +25,7 @@ class PathCmd(SqTableCommand):
         src: str = "",
         dest: str = "",
         vrf: str = "",
+        query_str: str = ' ',
     ) -> None:
         super().__init__(
             engine=engine,
@@ -34,6 +35,7 @@ class PathCmd(SqTableCommand):
             view=view,
             namespace=namespace,
             columns=columns,
+            query_str=query_str,
             format=format,
             sqobj=PathObj
         )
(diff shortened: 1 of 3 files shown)


suzieq/cli/sqcmds/PathCmd.py
```diff
@@ -25,6 +25,7 @@ def __init__(
         src: str = "",
         dest: str = "",
         vrf: str = "",
+        query_str: str = ' ',
```
claudiolor, line 28: Are you passing a string with an empty char on purpose or it is a mistake?

ddutt: All query_str have an blank char string. I don't recall why its so.

ddutt: Oh I remember why. One of the pandas' versions had an issue with the query-str if we didn't end with a blank. I don't know if this is still the case with the 1.3.4 version we're using.

suzieq/engines/pandas/path.py
```diff
@@ -1120,7 +1121,8 @@ def get(self, **kwargs) -> pd.DataFrame:
             # This occurs when a path traversal terminates due to an error such
             # as loop detected
             final_paths = paths
-        return self._path_cons_result(final_paths)
+        return self._handle_user_query_str(
+            self._path_cons_result(final_paths), query_str)
```
ddutt, line 1125: You need to add reset_index for this to work correctly.

ddutt (changes requested):

ddutt (approved):

PR #793, opened 2022-09-02
Fix device version ne

ddutt:
When version string in device command is given a ! such as "!9.3.7", it doesn' work. version expects "!=9.3.7". This is inconsistent with what all the other filters take for not equal to. 

To stay backwards compatible, I decided to continue to support !=. Added an additional test for !

### DIFF
--- a/suzieq/engines/pandas/device.py
+++ b/suzieq/engines/pandas/device.py
@@ -117,9 +117,11 @@ class DeviceObj(SqPandasEngine):
             df = df.loc[df.status.isin(status)]
         if os_version:
             opdict = {'>': operator.gt, '<': operator.lt, '>=': operator.ge,
-                      '<=': operator.le, '=': operator.eq, '!=': operator.ne}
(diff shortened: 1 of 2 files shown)


suzieq/engines/pandas/device.py
```diff
@@ -117,9 +117,10 @@ def get(self, **kwargs):
             df = df.loc[df.status.isin(status)]
         if os_version:
             opdict = {'>': operator.gt, '<': operator.lt, '>=': operator.ge,
-                      '<=': operator.le, '=': operator.eq, '!=': operator.ne}
+                      '<=': operator.le, '=': operator.eq, '!': operator.ne}
             op = operator.eq
             for osv in os_version:
+                osv = osv.replace('!=', '!')
```
claudiolor, line 123: Can you please add a comment to specify that we do this replacement for backward compatibility?

ddutt: Will do.

PR #794, opened 2022-09-02
Fix duplicates on result of filters and fix interval match and namespace filtering

claudiolor:
## Description

This PR fixes three issues in filtering:

Namespace filters has some problems due to the fact that we filter the files, and we directly applied regex on the filename, this caused some problems when for example the regex was looking for a suffix. `.*os` matched both:
```
namespace=nxos/
```
and 
```
namespace=my-ns/hostname=leaf-nxos
```
To fix it we just get the namespace part and filter on it.

A bug in the code caused the output of the show function to have some duplicates when multiple filters combined with OR matched the same rows. We fixes this issue by:
- Building a pandas query string and applying it once for hostname filters
- Removing the duplicates once we filter for the vlan filter in interface

Additionally, with PR #783 we added the possibility to filter by interval. The way we used to detect the interval has some problems, since it wasn't possible to look for extremes so ( i.e. X < 9 and X > 20). With this fix we detect an interval only if we find a sequence of >/>= and </<= and not if we just match them in any position.

Moreover, now we accept any combination of filters, both match and NOT filters are allowed:
i.e.:
```
hostname="~leaf.* !leaf01"
```

## Type of change

### DIFF
--- a/suzieq/db/parquet/parquetdb.py
+++ b/suzieq/db/parquet/parquetdb.py
@@ -8,6 +8,7 @@ from datetime import datetime, timedelta, timezone
 from contextlib import suppress
 from shutil import rmtree
 from collections import defaultdict
+import operator
 
 import pandas as pd
 import numpy as np
@@ -21,7 +22,6 @@ from suzieq.shared.schema import Schema, SchemaForTable
 from suzieq.db.parquet.pq_coalesce import (SqCoalesceState,
                                            coalesce_resource_table)
 from suzieq.db.parquet.migratedb import get_migrate_fn
-from suzieq.shared.utils import reduce_filter_list
 
 
 PARQUET_VERSION = '2.4'
@@ -639,7 +639,7 @@ class SqParquetDB(SqDB):
 
         return msch
 
-    def _get_filtered_fileset(self, dataset: ds, namespace: list) -> ds:
+    def _get_filtered_fileset(self, dataset: ds, namespaces: list) -> ds:
         """Filter the dataset based on the namespace
 
         We can use this method to filter out namespaces and hostnames based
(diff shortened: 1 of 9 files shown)


ddutt (changes requested):

suzieq/db/parquet/parquetdb.py
```diff
@@ -651,45 +651,65 @@ def _get_filtered_fileset(self, dataset: ds, namespace: list) -> ds:
         Returns:
             ds: pyarrow dataset of only the files that match filter
         """
-        if not namespace:
+        def check_ns_conds(ns_to_test: str, filter_list: List,
```
claudious96, line 654:
```suggestion
        def check_ns_conds(ns_to_test: str, filter_list: List[str],
```

suzieq/db/parquet/parquetdb.py
```diff
@@ -651,45 +651,65 @@ def _get_filtered_fileset(self, dataset: ds, namespace: list) -> ds:
...
+                           cond: operator.or_) -> bool:
+            # pylint: disable=comparison-with-callable
+            res = not cond == operator.or_
```
claudious96, line 657: What happens in this line? Also I don't understand why cond has that type hint

claudiolor: Yes, you are right maybe a comment or making it more explicit would help.
If I want to concat the expression with a OR i want the initial value to be False
while with AND I want the initial value to be True

suzieq/db/parquet/parquetdb.py
claudiolor: Fixed

claudious96 (approved):

suzieq/db/parquet/parquetdb.py
```diff
@@ -651,45 +651,65 @@ def _get_filtered_fileset(self, dataset: ds, namespace: list) -> ds:
...
-                    [file for file in chklist
-                     if (ns_section := next(
-                         (s for s in file.split('/')
-                          if s.startswith('namespace=')),
-                         None))
-                     and not re.search(f'namespace={ns_filter}', ns_section)]
-                # NOTs turn list from implicit OR into implicit AND
-                chklist = ns_filelist
+        for ns_match in namespaces:
+            if ns_match.startswith(('!~', '~!')):
```
LucaNicosia, line 686: `~!` is no more a valid filter. if the user types `~!`, the `!` must be considered the first character of the regex

suzieq/db/parquet/parquetdb.py
```diff
@@ -651,45 +651,65 @@ def _get_filtered_fileset(self, dataset: ds, namespace: list) -> ds:
...
+                not_filters.append(ns_match[2:])
```
LucaNicosia, line 687: why not escaping also this?

suzieq/db/parquet/parquetdb.py
```diff
@@ -651,45 +651,65 @@ def _get_filtered_fileset(self, dataset: ds, namespace: list) -> ds:
...
+            elif ns_match.startswith('~'):
+                match_filters.append(ns_match[1:])
+            elif ns_match.startswith('!'):
+                not_filters.append(re.escape(ns_match[1:]))
```
LucaNicosia, line 691: What appends if the list of namespaces do not contain any regex?

suzieq/shared/utils.py
```diff
@@ -569,78 +570,137 @@ def convert_numlist_to_ranges(numList: List[int]) -> str:
...
+                                    and f.startswith(('~', '!~'))):
+                continue
 
-    return query_str
+            for fval in filter_vals:
+                val_to_use = fval
+                use_not = False
+                regex = False
+                if isinstance(fval, str):
+                    if fval.startswith(('!~', '~!')):
```
LucaNicosia, line 663: same here for `~!`

suzieq/engines/pandas/interfaces.py
```diff
@@ -100,55 +100,85 @@ def _check_vlan_match(self, vlan_list: List[str],
...
+                elif expression.startswith('!'):
+                    val = expression[1:]
+                    op = expression[:1]
+            return op, val
+
+        if not vlan_filters:
+            return df
 
-        return df.query('vlanList.str.len() > 0 or vlan != 0')
+        opdict.update({'><': _interval})
```
LucaNicosia, line 134: why you update opdict instead of directly add this option at line 103?

suzieq/engines/pandas/interfaces.py
```diff
@@ -100,55 +100,85 @@ def _check_vlan_match(self, vlan_list: List[str],
...
+                not_filters.append(
+                    f'(vlan != {fval} '
+                    f'and ~@_check_cond(vlanList, "==", {fval}))')
+            elif (op.startswith('>')
+                  and i+1 < len(vlan_filters)
+                  and isinstance(vlan_filters[i+1], str)
+                  and vlan_filters[i+1].startswith('<')):
+                # In this case we are checking if the user asked for an
+                # an interval. So if we find a sequence of > and <,
+                # we will combine the rules
```
LucaNicosia, line 151: Just to be sure I understood correctly:
- `>10 <50` is equal to `10 < x < 50`
- `<50 >10` is equal to `x < 50 or x > 10` (which is everything)

Correct?

tests/integration/sqcmds/vmx-samples/swport.yml
```diff
@@ -44,5 +44,5 @@ tests:
 - command: interface show --columns=vlan --vlan='101 201' --namespace=vmx --format=json
   data-directory: tests/data/parquet/
   marks: interface show vmx switchport
-  output: '[{"vlan": 0}, {"vlan": 101}, {"vlan": 0}, {"vlan": 101}, {"vlan": 0}, {"vlan":
-    201}, {"vlan": 0}, {"vlan": 201}]'
+  output: '[{"vlan": 101}, {"vlan": 101}, {"vlan": 0}, {"vlan": 201}, {"vlan": 0},
+    {"vlan": 201}]'
```
LucaNicosia, line 48: why vlan `0` is still in the output?

suzieq/db/parquet/parquetdb.py
claudiolor: I call `re.escape` to match the exact string when the user doesn't explicitly ask for a regex pattern

suzieq/db/parquet/parquetdb.py
claudiolor: As before, we escape the string and we look for the exact match

suzieq/engines/pandas/interfaces.py
claudiolor: I know it's ugly but inside the _interval() function we use `opdict`, so when I define `opdict`,  `_interval()` has not been already defined

tests/integration/sqcmds/vmx-samples/swport.yml
claudiolor: Because we look to both `vlan` and `vlanList`, in this case vlanList contains `[101, 201]`

suzieq/engines/pandas/interfaces.py
claudiolor: Yes, at the moment we detect an interval only when the rule is written following the right order

ddutt (approved):

PR #795, opened 2022-09-09
sq-coalescer: fix old schemas handling and migration

claudiolor:
## Description

This PR fixes some issues with the coalescer when there are files following an old schema:

- Handle the case in which the coalescer reads file belonging to multiple schemas
- Introduce a "generic migration" function which is executed every time the coealscer converts data belonging to an old schema
- Remove dead code and code reorganization 

## Type of change

### DIFF
--- a/suzieq/db/parquet/migratedb.py
+++ b/suzieq/db/parquet/migratedb.py
@@ -53,21 +53,27 @@ def generic_migration(df: pd.DataFrame, table: str,
         df[column.name] = default_val
 
     # convert all dtypes to whatever is desired
-    for column in df.columns:
-        schema_type = arrow_schema.field(column).type.to_pandas_dtype()
-        if df.dtypes[column] != schema_type:
-            # when conversion can't be automatically performed we need to write
-            # a specific migration fn. Reraise the exception to detect this
-            # error
-            try:
-                df[column] = df[column].astype(schema_type)
-            except ValueError as e:
-                raise ValueError(
-                    f'Unable to perform auto conversion of {table} '
-                    f'to {schema.version}, due to: {e}')
+    columns_in_schema = arrow_schema.names
+    for column in columns_in_schema:
+        # If the column was missing, there is no need to try conversion, since
+        # there is no previous type
+        if column not in missing_fields:
+            schema_type = arrow_schema.field(column).type.to_pandas_dtype()
+            if df.dtypes[column] != schema_type:
+                # when conversion can't be automatically performed we need to
+                # write a specific migration fn. Reraise the exception to
+                # detect this error
+                try:
+                    df[column] = df[column].astype(schema_type)
+                except ValueError as e:
+                    raise ValueError(
+                        f'Unable to perform auto conversion of {table} '
+                        f'to {schema.version}, due to: {e}')
     df['sqvers'] = schema.version
 
-    return df
+    # We would like to return only what is in the current schema, dropping all
+    # the old columns
+    return df[columns_in_schema]
 
 
 def _convert_bgp_vers_1_to_2(df: pd.DataFrame, **_) -> pd.DataFrame:
--- a/suzieq/db/parquet/migratedb.py
+++ b/suzieq/db/parquet/migratedb.py
@@ -2,6 +2,7 @@
 from typing import Callable, Union
 import pandas as pd
 from pandas.core.computation.ops import UndefinedVariableError
+from suzieq.shared.exceptions import SqVersConversionError
 from suzieq.shared.schema import Schema
 
 from suzieq.shared.utils import get_default_per_vals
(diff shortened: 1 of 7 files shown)


suzieq/db/parquet/pq_coalesce.py
```diff
@@ -1,17 +1,18 @@
...
+from collections import defaultdict
 from datetime import datetime, timedelta, timezone
-import logging
-from typing import List
 from itertools import repeat
-import tarfile
+from typing import List
 
 import pandas as pd
+import pyarrow.dataset as ds  # put this later due to some numpy dependency
```
claudious96, line 11: If everything works as expected, this comment is outdated and would be better to delete it.

suzieq/db/parquet/pq_coalesce.py
```diff
@@ -94,10 +95,21 @@ def write_files(table: str, filelist: List[str], in_basedir: str,
     state.block_start = int(block_start.timestamp())
     state.block_end = int(block_end.timestamp())
     if filelist:
-        this_df = ds.dataset(source=filelist, partitioning='hive',
-                             partition_base_dir=in_basedir) \
-            .to_table() \
-            .to_pandas()
+        files_pervers = defaultdict(list)
+        for file in filelist:
+            sqversmatch = re.search(r'sqvers=([0-9]*.[0-9]*)', file)
```
claudious96, line 100: The pattern is quite loose, if we are only expecting numeric versions (which I believe is our case) we can just change it to:
```suggestion
            sqversmatch = re.search(r'sqvers=([0-9]+.[0-9]+)', file)
```

LucaNicosia, line 100: The correct regex should be `sqvers=([0-9]+/.[0-9]+)`. You need to escape the `.` otherwise it will match any character

suzieq/db/parquet/parquetdb.py
```diff
@@ -389,64 +389,48 @@ def migrate(self, table_name: str, schema: SchemaForTable) -> None:
...
+                                        partitioning='hive') \
+                            .to_table() \
+                            .to_pandas(self_destruct=True)
+                    except pa.ArrowInvalid as e:
+                        self.logger.warning(f'Unable to migrate file: {e}')
+
+                    if migrate_rtn:
+                        df = migrate_rtn(df, table_name, schema)
+                    # Always perform generic migration
+                    df = generic_migration(df, table_name, schema)
```
LucaNicosia, line 415: Can you explain why always generic migration?

suzieq/db/parquet/migratedb.py
```diff
@@ -22,7 +25,49 @@ def get_migrate_fn(table_name: str, from_vers: str,
...
+        df[column.name] = default_val
+
+    # convert all dtypes to whatever is desired
+    for column in df.columns:
+        schema_type = arrow_schema.field(column).type.to_pandas_dtype()
+        if df.dtypes[column] != schema_type:
+            # If we are not able to perform the conversion, ignore and go on.
+            # We should not have this case, as when conversion can't be
+            # automatically performed we need to write a specific
+            # migration fn.
```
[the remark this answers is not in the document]
claudiolor: Probably ignoring is not even the right thing to do, we should be able to detect this kind of errors. Moreover, if we fail conversion here, even pyarrow won't be able to write the output file.

suzieq/db/parquet/parquetdb.py
claudiolor: This is to make things more robust. The generic migration allows us to always make sure that all the fields have the right type and if it is not the case, we try the conversion. I think there could be two different approaches:
- Let the specific migration function handle everything, so we run the generic_migration only when we don't have a migration function to run
- Simple conversions might not be handled by the specific migration functions, so we always the the generic one

LucaNicosia: I think the correct approach would be to keep the code as is, otherwise we need to take care of every possible problems in the specific migration function

suzieq/db/parquet/pq_coalesce.py
```diff
@@ -97,7 +97,7 @@ def write_files(table: str, filelist: List[str], in_basedir: str,
     if filelist:
         files_pervers = defaultdict(list)
         for file in filelist:
-            sqversmatch = re.search(r'sqvers=([0-9]*.[0-9]*)', file)
+            sqversmatch = re.search(r'sqvers=([0-9]+\.[0-9]+)', file)
```
ddutt, line 100: So, this will automatically convert OSS into enterprise if the user switches?

ddutt (changes requested):

suzieq/db/parquet/migratedb.py
```diff
@@ -56,12 +56,15 @@ def generic_migration(df: pd.DataFrame, _table: str,
...
-
+            # when conversion can't be automatically performed we need to write
+            # a specific migration fn. Reraise the exception to detect this
+            # error
+            try:
+                df[column] = df[column].astype(schema_type)
+            except ValueError as e:
+                raise ValueError(
+                    f'Unable to perform auto conversion of {table} '
+                    f'to {schema.version}, due to: {e}')
```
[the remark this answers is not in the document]
claudiolor: Yes, we abort the coalescing of the specific table

suzieq/db/parquet/migratedb.py
```diff
@@ -22,7 +25,49 @@ def get_migrate_fn(table_name: str, from_vers: str,
...
+    use the generic one.
+
+    Args:
+        df (pd.DataFrame): DataFrame to migrate
+        schema (Schema): the current schema
+
+    Returns:
+        pd.DataFrame: the migrated DataFrame
+    """
+
```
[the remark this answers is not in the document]
claudiolor: Sure

suzieq/db/parquet/migratedb.py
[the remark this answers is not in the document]
ddutt: We shouldn't just abort the coalescer for the specific table. This is potentially a bad situation that needs the attention of the user ASAP. Maybe abort the poller ?

claudious96 (approved):

suzieq/db/parquet/migratedb.py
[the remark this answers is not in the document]
claudiolor: Fixed

ddutt (approved):

PR #796, opened 2022-09-17
Set vrf to dvrf if port is trunk

tgupta3:
## Description
In case of vxlan, it's very common to have trunk ports facing towards the servers. I observed if the devices_iifs is `{'iif': 'Ethernet1/26', 'mtu': 9216, 'outMtu': 9216, 'overlay': '', 'protocol': '', 'hopError': [], 'lookup': '1.1.1.1', 'macaddr': None, 'vrf': 'vrf1', 'mtuMatch': True, 'is_l2': False, 'nhip': '', 'overlay_nhip': '', 'oif': 'Ethernet1/26', 'timestamp': Timestamp('2022-09-16 16:13:09.461000-0700', tz='America/Los_Angeles'), 'l3_visited_devices': set(), 'l2_visited_devices': set()}`, the vrf was being set to native vlan which causes issue in routing path. 

The fix will just set the vrf to empty so that it gets overwritten to dvrf in https://github.com/netenglabs/suzieq/blob/aa4fbd6baeb20e2fb912d9cf2414810b9612d5f3/suzieq/engines/pandas/path.py#L968

## Type of change

## New Behavior

The source will no longer get attributed to native vlan vrf .

Set vrf to vrf of the ip if the port is trunk

### DIFF
--- a/suzieq/engines/pandas/path.py
+++ b/suzieq/engines/pandas/path.py
@@ -42,7 +42,8 @@ class PathObj(SqPandasEngine):
                 .get(namespace=namespace, state='up',
                      columns=['namespace', 'hostname', 'ifname', 'mtu', 'type',
                               'ipAddressList', 'ip6AddressList', 'state',
-                              'vlan', 'master', 'macaddr', 'timestamp', 'portmode']) \
+                              'vlan', 'master', 'macaddr', 'timestamp',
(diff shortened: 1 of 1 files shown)


ddutt (approved):

ddutt:
Thanks for the patch @tgupta3

PR #797, opened 2022-09-19
path: fix reverse path check logic when nodes are evpn

ddutt:
The issue in this case is in the logic for checking if there's a reverse path. When the path starts via the ARP table on a leaf, its possible that its a pure L2 EVPN network and as a consequence, they have no routing table entry for the tunnelled IP address. Instead of failing the path, we need to check if the ARP table has an entry for that IP address, and if not warn that there could be a flood if sent to that address. This avoids false "no reverse path" error messages. 


suzieq/engines/pandas/path.py
```diff
@@ -853,7 +853,15 @@ def get(self, **kwargs) -> pd.DataFrame:
                         vrfchk = dest_device_iifs[destdevkey]["vrf"]
                         rev_df = self._rpf_df.query(
                             f'hostname == "{device}" and vrf == "{vrfchk}"')
-                        if rev_df.empty:
+                        if rev_df.empty and is_l2:
+                            # for the case if this node is the ingress VTEP
+                            rev_df = self._arpnd_df.query(
+                                f'hostname == "{device}" and '
+                                f'ipAddress == "{dest}"')
+                            if rev_df.empty and is_l2:
```
LucaNicosia, line 861: If we entered here, `is_l2` is already True. We can simplify the check as follows
```suggestion
                            if rev_df.empty:
```

LucaNicosia (commented):
Is it possible to add tests for this?

ddutt:
Adding tests is possible but requires me to modify dual-evpn to get rid of the servers

PR #799, opened 2022-09-20
sq-poller: fix issues with device connection

claudiolor:
## Description

This PR addresses a set of issues related to the connection with the poller devices:
- ios, iosxe and iosxr: fix initial connection
- ios, iosxe and iosxr: do not disconnect after discovery if devtype is provided
- ios, iosxe: if slow_host is provided increase the initial connection backoff time to 30
- all: do not call device data initialization if connection fails
- ios, iosxe, iosxr: fix deadlock happening in some corner cases
- all: fix max recursion error happening in some corner cases while retrieving the device data 

## Type of change

### DIFF
--- a/suzieq/poller/worker/nodes/node.py
+++ b/suzieq/poller/worker/nodes/node.py
@@ -688,7 +688,8 @@ class Node:
             # Release here because init_dev_data uses this lock as well
             # We need to be sure that devtype is set, otherwise the
             # _fetch_dev_data function is not implemented and will raise.
-            if init_dev_data and self.devtype:
+            # Moreover we want to be connected if we call this function
+            if self._conn and init_dev_data and self.devtype:
                 await self._fetch_init_dev_data()
 
     @abstractmethod
--- a/suzieq/poller/worker/nodes/node.py
+++ b/suzieq/poller/worker/nodes/node.py
@@ -1435,13 +1435,16 @@ class IosXRNode(Node):
                     stderr=DEVNULL)
                 self.logger.info(
                     f'Persistent SSH present for {self.hostname}')
-                if init_dev_data:
-                    await self._fetch_init_dev_data()
             except Exception:
                 self._conn = self._long_proc = None
 
         if use_lock:
             self.ssh_ready.release()
+
+        # We need to release the lock and make sure that we are connected to
+        # the device before proceeding with the _fetch_init_dev_data() fn.
+        if self.is_connected and init_dev_data:
+            await self._fetch_init_dev_data()
 
     async def _parse_init_dev_data(self, output, cb_token) -> None:
         '''Parse the version for uptime and hostname'''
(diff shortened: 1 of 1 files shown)


suzieq/poller/worker/nodes/node.py
```diff
@@ -214,8 +214,12 @@ async def initialize(self, **kwargs) -> TNode:
             # are executed. So close the conn at our end too
             # avoiding the initial persistent connection failure that
             # otherwise happens
-            if self.devtype in ['iosxe', 'ios', 'iosxr']:
+            if self.devtype in ['iosxe', 'ios', 'iosxr'] and self._conn:
                 await self._close_connection()
+                # In this case opening and closing a connetion too quickly
+                # might cause an authentication failure. Wait some time before
+                # proceeding with the new connection
+                await asyncio.sleep(60)
```
ddutt, line 223: Can we define 60 as a constant please?

suzieq/poller/worker/nodes/node.py
```diff
@@ -1490,6 +1490,11 @@ async def _init_ssh(self, init_dev_data=True,
         authenticated, not just the main SSH connection.
         '''
         backoff_period = 1
+        # If the device is marked as slow, we don't want to be too fast when
+        # we retry the connection. So increase the initial backoff time to 30s.
+        if self.slow_host:
+            backoff_period = 30
+
```
ddutt, line 1531: Why not do an if-else here? if slow_host 30 else 1?

suzieq/poller/worker/nodes/node.py
```diff
@@ -1490,6 +1490,11 @@ async def _init_ssh(self, init_dev_data=True,
         authenticated, not just the main SSH connection.
         '''
         backoff_period = 1
```
ddutt, line 1526: Can we also fix the commit message please? We're increasing connection backoff time, not reducing it

suzieq/poller/worker/nodes/node.py
```diff
@@ -947,10 +950,29 @@ async def _exec_service(self, service_callback, svc_defn: dict,
...
+        # the device data, so if we establish a new connection, we should not
+        # call that function again
+        self._fetching_dev_data = True
+        try:
+            await self._fetch_init_dev_data_internal()
+        finally:
+            self._fetching_dev_data = False
+
+    @abstractmethod
+    async def _fetch_init_dev_data_internal(self):
```
ddutt, line 973: Instead of internal, is the name fetch_init_dev_data_devtype better?

ddutt (changes requested):

suzieq/poller/worker/nodes/node.py
```diff
@@ -28,6 +28,11 @@
 
 logger = logging.getLogger(__name__)
 IOS_SLEEP_BET_CMDS = 5          # in seconds
+IOS_TIME_AFTER_DISCOVERY = 60   # time to wait after ios(xe,xr) auto-discovery
+IOS_CONN_INITIAL_BACKOFF = 1
+IOS_CONN_INITIAL_BACKOFF_SLOW = 30
+IOS_CONNECTION_ATTEMPTS = 4
```
FloLaco, line 34: Why not using a parameter from config file or inventory and set default value here ?

ddutt: We're trying to limit the number of configurable knobs needed. Once you put one out, its hard to take it back.

LucaNicosia (approved):

ddutt (approved):

PR #821, opened 2023-01-09
Initial Junos EVO support

claudiolor:
## Description

This PR introduces initial support for Junos EVO devices. The following services are available:

- aprnd
- bgp
- devconfig
- device
- evpnVni
- interfaces
- inventory
- lldp
- macs
- ospfIf
- ospfNbr
- routes
- vlan

## Type of change

### DIFF
--- a/suzieq/poller/worker/nodes/node.py
+++ b/suzieq/poller/worker/nodes/node.py
@@ -366,6 +366,11 @@ class Node:
                         devtype = 'junos-es'
                 if not devtype:
                     devtype = "junos"
+            elif 'Junos: ' in data:
+                if re.search(r'-EVO\b', data):
+                    devtype = 'junos-evo'
+                else:
+                    devtype = 'junos'
             elif "NX-OS" in data:
                 devtype = "nxos"
             elif "SONiC" in data:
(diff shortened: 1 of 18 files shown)


suzieq/poller/worker/nodes/node.py
```diff
@@ -1892,6 +1902,10 @@ async def _parse_init_dev_data_devtype(self, output, cb_token) -> None:
             data = output[1]["data"]
             hmatch = re.search(r'\nHostname:\s+(\S+)\n', data)
             if hmatch:
+                # TODO
```
claudiolor, line 1905: Should we need to address this TODO now?

suzieq/poller/worker/nodes/node.py
```diff
@@ -1863,18 +1868,23 @@ async def _rest_gather(self, service_callback, cmd_list, cb_token,
...
-        await self._exec_cmd(self._parse_init_dev_data,
-                             ["show system uptime|display json",
-                              "show version"], None, 'mixed',
-                             reconnect=reconnect)
+        if self.devtype == 'junos-evo':
+            cmdlist = ["show system uptime|display json",
+                       'show version node all']
+        else:
+            cmdlist = ["show system uptime|display json",
+                       'show version']
```
LucaNicosia, line 1876: @ddutt why don't create a new class for junos evo instad of doing this if else? It seems a cleaner solution

ddutt: Because we'll be duplicating a lot of work

suzieq/poller/worker/nodes/node.py
ddutt: Not right now.

ddutt (approved):

ddutt (approved):

PR #823, opened 2023-01-09
sq-poller: add reboot detection

claudiolor:
## Description

This PR adds back reboot detection in the poller, allowing it to check whether one of the devices rebooted and in this case it will rewrite all the records, this allows the user to see changes whenever a device reboots.

Additionally the PR tries to reduce the time drift of the bootup timestamp due to the fact that some of the devices return the uptime instead of a epoch timestamp, therefore we need to work out the timestamp. To mitigate this issue we:

- Always use the response timestamp to work out the bootup timestamps
- Always use seconds for the bootup timestamps
- Get the epoch timestamp instead of uptime in junos, linux, cumulus and sonic

## Type of change


tests/integration/test_schema_read.py
```diff
@@ -45,23 +45,12 @@ def test_schema_data_consistency(table, datadir, columns, get_table_data_cols):
     assert not df.empty
 
     sqobj = get_sqobject(table)()
```
LucaNicosia, line 47: This sqobj is initialized with the default config file. If it's not present, like in git tests, this sqobject will fail.

``` suggestion
    config_file = create_dummy_config_file()
    sqobj = get_sqobject(table)(config_file=config_file)
```

claudious96 (approved):

suzieq/config/schema/address.avsc
```diff
@@ -65,7 +65,7 @@
         {
             "name": "vrf",
             "type": "string",
-	    "depends": "master",
+            "depends": "master",
```
ddutt, line 68: I don't understand why we have the commits on the schema file in this patch. All these seem like whitespace fixes and they're just different from what was there. We should remove these from this patchset. This obviously excludes the deviceSession field addition

suzieq/poller/worker/services/service.py
```diff
@@ -26,6 +26,8 @@
 HOLD_TIME_IN_MSECS = 60000
 # How many unsuccessful polls before marking records with active=false
 HYSTERESIS_FAILURE_CNT = 3
+# The boot timestamp time drift tolerance (in seconds)
+TIME_DRIFT_TOLERANCE = 30
```
ddutt, line 30: 30s seems a lot. Why this high? This assumes a default polling period of 1 min?

ddutt (changes requested):

suzieq/poller/worker/services/service.py
claudiolor: The polling period is not really involved, since the boot timestamp is computed when we initialize the the device data, after we connect (or reconnect) to the device. The aim of the time drift is compensating the time required for the request to arrive to the device, which is variable. 30 is a high value, the aim was compensating a possible delay in the network, but at the same time probably it's too much and we might miss a reboot. So, if you have a better value, happy to change it

suzieq/config/schema/address.avsc
claudiolor: In some places in the schema there were tabs instead of the 4 spaces. The editor changed them once I saved. I kept spaces on purpose and I recommend to keep them because every time someone changes the schema that tabs will get replaced.

PR #842, opened 2023-01-18
sq-poller: remove rest implementation for Cumulus

claudiolor:
## Description

Even though the documentation says Cumulus didn't support REST, the Cumulus Node class actually had a not working implementation of the REST communication. This PR removes it, as dead code.

## Type of change

### DIFF
--- a/suzieq/poller/worker/nodes/node.py
+++ b/suzieq/poller/worker/nodes/node.py
@@ -1516,6 +1516,15 @@ class EosNode(Node):
 class CumulusNode(Node):
     '''Cumulus Node specific implementation'''
 
+    async def _rest_connect(self):
+        raise NotImplementedError(
+            f'{self.address}: REST transport is not supported')
+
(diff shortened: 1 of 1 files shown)


suzieq/poller/worker/nodes/node.py
```diff
@@ -1568,6 +1568,15 @@ def _extract_nos_version(self, data: str) -> None:
 class LinuxNode(CumulusNode):
     '''Linux server node'''
 
+    async def _rest_connect(self):
```
ddutt, line 1571: Since its inheriting from CumulusNode which already rejected REST, do we need this too? Or is this just additional protection?

claudiolor: In theory it shouldn't be needed but If we don't override the methods also in the LinuxNode pylint complains, I believe because since the methods in Cumulus raise a NotImplementedError pylint still considers the methods as abstract

ddutt: Thanks

PR #851, opened 2023-01-31
Poller: handle multiple exceptions

LucaNicosia:
## Description

The entire code of the controller and the workers leverage on asyncio tasks.

In the current version of the code, if one of the task was raising an exception, we fail right away and re-raise the exception letting the core method of the SuzieQ components handle the exception.
It may happen that more than one task raise an exception: in this case, since we raise an exception as soon as we see one, the other exceptions are not handled. 
For these exceptions, asyncio complains with the message `Task exception was never retrieved` (see [python docs](https://docs.python.org/3/library/asyncio-dev.html#detect-never-retrieved-exceptions)) and re-raise the exceptions, but since the main code of the SuzieQ component is already ended, the exception is not handled and the traceback is printed on the log generating a confusing output.

To fix this issue, it is introduced a new exception class called `SqRuntimeError` which contains a list of exceptions.
In the code of the controller and the worker, instead of raising the exception as soon as we see a task failing, we collect all the exception in an array. If this array is not empty, we use this array to initialize a `SqRuntimeError` exception and we raise this exception.
The main code of the controller (sq_poller.py) and the worker (sq_worker.py) will verify if the type of the exception is `SqRuntimeError` and if so, log each exception in the exception list.

## Type of change

## Double Check

### DIFF
--- a/suzieq/poller/controller/controller.py
+++ b/suzieq/poller/controller/controller.py
@@ -13,15 +13,16 @@ import asyncio
 import logging
 import signal
 from collections import defaultdict
+from copy import deepcopy
 from pathlib import Path
 from typing import Dict, List
-from copy import deepcopy
 
 from suzieq.poller.controller.base_controller_plugin import ControllerPlugin
 from suzieq.poller.controller.inventory_async_plugin import \
     InventoryAsyncPlugin
 from suzieq.poller.worker.services.service_manager import ServiceManager
-from suzieq.shared.exceptions import InventorySourceError, SqPollerConfError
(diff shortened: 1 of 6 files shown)


claudious96 (approved):

claudiolor (approved):

ryanmerolle:
Can this be merged?

ddutt:
> Can this be merged?

Yes, we'll work on getting this done for a possible end-of-month release

ryanmerolle:
Merging this fix in helps tackle a lot of my current poller issues.  I tested in my fork and everything looks great. How can I help here?

PR #853, opened 2023-02-02
Update docs 0.20.0

ddutt:
Updating the docs. Many involve just changing Suzieq to SuzieQ, but there are significant updates otherwise.


docs/poller.md
```diff
@@ -44,34 +44,44 @@ The poller could use different policies for the inventory splitting. At the mome
...
 
 ```yaml
 poller:
   chunker:
      policy: sequential
 ```
 
+### How Many Workers??
+
+An often asked question is how many workers are necessary for any given network. The answer depends on many factors such as the size of the data being pulled, the frequency of the polling, how long each device takes to respond and so on. Some users with a powerful CPU (16 cores, newer processor etc.) have been using a single worker with upto 600 devices while others have had to run a worker per 30 or so devices for the same polling interval.
```
claudious96, line 57:
```suggestion
An often asked question is how many workers are necessary for any given network. The answer depends on many factors such as the size of the data being pulled, the frequency of the polling, how long each device takes to respond and so on. Some users with a powerful CPU (16 cores, newer processor etc.) have been using a single worker with up to 600 devices while others have had to run a worker per 30 or so devices for the same polling interval. 
```

docs/poller.md
```diff
@@ -44,34 +44,44 @@ The poller could use different policies for the inventory splitting. At the mome
...
... sq-poller -I inventory.yml --run-once=update``` and assuming no user input is necessary (such as typing a password), the output reveals roughly how long it takes to discover, and gather the data from all of the devices just once. An examination of the totalTime per service reveals what a normal polling interval ought to be. Also look at the device's CPU utilization to see the cost of observing the data just once. You can always alter the polling frequency of each service by modifying the period value (or adding a key-value such as ``period: 300`` to the service config file. These files are typically found under the config directory of wherever suzieq happens to be installed. See the devconfig.yml file under the config directory for an example of adding a period. You can also run multiple pollers each with a different inventory (be very careful to not mix up the device list) and different suzieq-cfg.yml files, each containing a different default polling period, but using the same data-directory to ensure you have all the data in one place. For example, you can put all your JunOS devices in one inventory and monitor it with a different polling period compared to say Cumulus devices.
```
claudious96, line 63: A closing parenthesis was missing.
```suggestion
**A quick way to get a measure of things** is to run the poller in the snapshot mode (use ```--run-once=update``` option with the poller) where it gathers the data just once and terminates. For example, with an inventory file called inventory.yml, if you launch the poller as ```time sq-poller -I inventory.yml --run-once=update``` and assuming no user input is necessary (such as typing a password), the output reveals roughly how long it takes to discover, and gather the data from all of the devices just once. An examination of the totalTime per service reveals what a normal polling interval ought to be. Also look at the device's CPU utilization to see the cost of observing the data just once. You can always alter the polling frequency of each service by modifying the period value (or adding a key-value such as ``period: 300`` to the service config file). These files are typically found under the config directory of wherever suzieq happens to be installed. See the devconfig.yml file under the config directory for an example of adding a period. You can also run multiple pollers each with a different inventory (be very careful to not mix up the device list) and different suzieq-cfg.yml files, each containing a different default polling period, but using the same data-directory to ensure you have all the data in one place. For example, you can put all your JunOS devices in one inventory and monitor it with a different polling period compared to say Cumulus devices.

```

claudious96 (approved):
Just a few typos

claudious96:
@ddutt please remove the additional newlines at the end of `docs/tutorials/how_to_think_about_verbs.md`, pre-commit is failing because of that.

LucaNicosia (commented):
If possible, can you change in the bash examples the `ddutt>` to `user>`? It's more generic

ddutt:
> If possible, can you change in the bash examples the `ddutt>` to `user>`? It's more generic

Done. Replaced with suzieq>

ddutt:
> @ddutt please remove the additional newlines at the end of `docs/tutorials/how_to_think_about_verbs.md`, pre-commit is failing because of that.

Done

docs/poller.md
ddutt: Fixed

docs/poller.md
ddutt: Fixed

PR #882, opened 2023-05-08
Poller status normalize

ddutt:
This` patch introduces a stringified version of the status in sqPoller. Instead of showing status codes such as 127, 200 etc., it shows strings such as "Command Not Found", "OK" etc. To this end, we've replaced the column status with an augmented column, statusStr, that contains the status string. Command filters such as "status=fail" work as before.

In addition, query-str didn't work correctly for sqPoller. This patch makes it function like the other commands.

all sqPoller tests have been updated, including sqPoller describe.

The change is backwards-compatible, except for the bug fix to query-str which makes it function like the other commands.

sqPoller show now shows a status string instead of a numeric code, making it easier for users to understand what the error is in polling.


suzieq/shared/poller_error_codes.py
```diff
@@ -0,0 +1,60 @@
...
+    '''SuzieQ Poller Status Codes and their strings'''
+
+    OK = 0, 'OK'
+    HTTP_OK = 200, 'OK'
+    PERMISSION_DENIED = 1001, 'Permission Denied'
+    UNAUTHORIZED = 1002, 'Unauthorized'
+    LOGIN_FAIL = 1003, 'Login Failed'
+    PRIV_ESC_FAIL = 1004, 'Privilege Escalation Failed'
+    NON_FINAL_FAILURES = 1005, 'Dummy to mark boundary, never set'
+    CMD_NOT_FOUND = 1006, 'Command Not Found'
```
claudious96, line 18: Can we use 2000+ for non final failures?

ddutt: Done

claudious96 (approved):

PR #886, opened 2023-05-09
Device namespace fixes

ddutt:
## Description

Three somewhat related bug fixes:
* namespace table: user query was applied too early resulting in either query failures or the wrong data being filtered
* namespace table: the schema didn't contain the lastUpdate column
* device: the filters such as os, vendor etc didn't work correctly because of an incorrect merge with the poller table. There's still a corner case where it fails. 

## Type of change

These are bug fixes. One test was updated to handle the changes caused by these fixes.

## New Behavior

namespace show honors the user filters plus query-str correctly.

* namespace's schema reflects the actual namespace columns
* namespace honors the filters the user specifies
* device honors the os/vendor/model filters the user specifies.

### DIFF
--- a/suzieq/engines/pandas/device.py
+++ b/suzieq/engines/pandas/device.py
@@ -52,12 +52,15 @@ class DeviceObj(SqPandasEngine):
         if view == 'latest' and 'status' in df.columns:
             df['status'] = np.where(df.active, df['status'], 'dead')
 
-        poller_df = pd.DataFrame()
+        poller_df: pd.DataFrame = pd.DataFrame()
+
+        # Retrieve the devices that have possibly not made it into the device
+        # list because they never could connect due to any reason
         if not ignore_neverpoll:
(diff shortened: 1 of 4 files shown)


suzieq/engines/pandas/device.py
```diff
@@ -77,7 +80,7 @@ def get(self, **kwargs):
                 .reset_index(drop=True)
 
             df = df.merge(poller_df, on=['namespace', 'hostname'],
-                          how='outer', suffixes=['', '_y'])  \
+                          how='outer', suffixes=['', '_y'], indicator=True) \
```
LucaNicosia, line 83: I think you added the indicator only for debug. It's not used in the next code lines. We should remove it.

Since you are working on these lines, can you fix the pyright problem as follows?
``` suggestion
                          how='outer', suffixes=('', '_y')) \
```

ddutt: Done

Issue #875, opened 2023-04-02
[Docs]: Transport setting is ignored for Arista EOS devices listed in Ansible inventory


ipspace:
### Change Type

Correction

### Area

Configuration parameters

### Proposed Changes

The Ansible section of Inventory File document (https://suzieq.readthedocs.io/en/latest/inventory/#ansible) mentions REST transport as the assumed transport for Arista EOS devices specified in an Ansible inventory, but does not mention that you cannot override that setting with `transport` parameter of `devices` section.

For example, `sq-poller` still uses REST API to collect data from Arista EOS devices when using the following configuration file:

```
---
sources:
- name: netlab-ansible
  type: ansible
  path: /cfdir/ansible-inventory.json

devices:
- name: ignore-known-hosts
  ignore-known-hosts: true
  transport: ssh

auths:
- name: vagrant
  username: vagrant
  password: plain:vagrant

namespaces:
- name: netlab
  source: netlab-ansible
  device: ignore-known-hosts
  auth: vagrant
```

ddutt:
We'll fiix this in the code.

ddutt:
Fixed

PR #887, opened 2023-05-09
Fix 875

ddutt:
## Description

Due to a quirk of my testing, I had forced the transport for EOS devices to always be REST. This obviously doesn't work in all situations, as was reported in issue #875 . This patch fixes the code.

## Type of change

Bug fix and documentation update.

## New Behavior

EOS hosts will assume the transport is ssh unless otherwise specified in the device section of the inventory file. 

EOS hosts were assumed to always be contacted via the REST API independent of what the device section of the SuzeieQ inventory file said.

Not everyone turns on REST API, and if those users used Ansible source, SuzieQ would be unable to poll those devices. Now, the user can specify the transport via the device section in the SuzieQ inventory file, subject to the caveat that with Ansible source, the device and auth sections cannot be specified per host or group of hosts..

updated docs included in this patch.

BREAKING CHANGE: With Ansible sources in the inventory file, ssh is assumed to be the default transport, including for EOS devices for which prior versions had forced the version to be REST.


LucaNicosia (approved):

suzieq/poller/controller/source/ansible.py
```diff
@@ -108,15 +108,7 @@ def _get_inventory(self) -> Dict:
...
 
-            # Retrieve password information
-            devtype = None
-            if entry.get('ansible_network_os') in ['eos', 'panos']:
-                devtype = entry.get('ansible_network_os')
-                transport = 'https'
-                port = 443
-            else:
-                transport = 'ssh'
-                port = entry.get('ansible_port', 22)
```
claudious96, line 119: If we remove this we won't be able to use the ansible inventories with panos devices. Can we leave it as is but checking only for `ansible_network_os` == `panos`?

ddutt: I had removed it for other reasons, but I think I agree with you. Adding it back in for panos.

claudious96 (approved):

LucaNicosia (approved):

Issue #865, opened 2023-03-06
[Bug]: when transport https set, defaults to use port 22


ryanmerolle:
### Suzieq version

0.20.1

### Install Type

container

### Python version

3.10

### Impacted component

sq-poller

### Steps to Reproduce

If you uncomment out the port to set it explicitly, everything works.

**inventory.yml**
```yaml
---
sources:
  - name: netbox
    token: TESTTOKEN
    url: https://netbox.example.com/
    tag:
      - test
    type: netbox
    period: 3600

devices:
  - name: devices-using-rest
    transport: https
    #port: 443
    devtype: eos
    ignore-known-hosts: true

auths:
  - name: TACACS
    username: test-username
    password: test-password

namespaces:
  - name: test
    source: netbox
    device: devices-using-rest
    auth: TACACS
```

**Log**
```
suzieq_poller        | [WORKER 0]: 2023-03-06 21:16:27,427 - suzieq.poller.worker.nodes.node - INFO - 192.168.1.1:22 exec: ['show version', 'show hostname']
```

### Expected Behavior

If https transport set, the port should default to 443.

### Observed Behavior

If https transport set, the port defaults to 22.

PR #897, opened 2023-05-30
bug fix on inventory set_device 

AndryNick98:
## Related Issue

#865

## Description

Fix the problem that occur when inside an inventory file be set https as transport but the port remain 22 instead of 443.
Modified set_device method inside source class to set the correct default port based on transport.
Fix mismanagement about ignore_known_hosts property.
Updated test.

## Type of change

## Double Check

### DIFF
--- a/tests/unit/poller/controller/sources/test_devices.py
+++ b/tests/unit/poller/controller/sources/test_devices.py
@@ -9,14 +9,14 @@ from suzieq.poller.controller.utils.inventory_utils import DeviceModel
 from suzieq.shared.utils import PollerTransport
 
 
-_INVENTORY = [{
+_INVENTORY = {
     'native-ns.192.168.123.123.443':
     {
         'address': '192.168.123.123',
         'hostname': None,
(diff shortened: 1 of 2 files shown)


tests/unit/poller/controller/sources/test_devices.py
```diff
@@ -80,7 +134,7 @@ async def test_devices_set(inventory: Dict):
 
         # emulate what the function Source.set_device should do
         exp_inv = {}
-        for key, node in inventory.items():
+        for key, node in result_inventory.items():
```
LucaNicosia, line 137: Your fix is absolutely correct, but now you don't need to compare one element at a time. You can simply compare `inv` with `result_inventory`
```suggestion
assert inv == result_inventory, 'inventories do not match'
```

LucaNicosia (commented):
There is a linting error to fix

claudious96 (approved):

ddutt (approved):

PR #909, opened 2023-07-15
Path improv

ddutt:
Handle path when the source is not in ARP or as interface on a polled node, by seeking the SVI associated with that source.

## Type of change

Enhancement

## New Behavior

Without this patch, if you provide a source IP address that is not in the arpnd table or as the address of an interface on a polled device, the command fails with "invalid src:...". With this patch, we find the SVI associated with the IP and start tracing from that node's SVI. 

If we cannot find an SVI, we flag the error: "unable to find a source node to start for <source IP>"

This is a useful feature for many people using path because the source's ARP entry may have timed out, people are testing a hypothetical source etc.

None.

ENHANCEMENT: path now supports specifying a source IP that may not be in the ARPND table. We determine the SVI node(s) associated with the IP and start the trace from there. Furthermore, the error message associated with a source IP whose source node we cannot locate is made more user-friendly now.

## Double Check

### DIFF
--- a/suzieq/engines/pandas/path.py
+++ b/suzieq/engines/pandas/path.py
@@ -1,3 +1,4 @@
+from typing import Optional, List, Any, Iterable
 from ipaddress import ip_network, ip_address
 from collections import OrderedDict
 from itertools import repeat
@@ -19,7 +20,7 @@ class PathObj(SqPandasEngine):
     '''Backend class to handle manipulating virtual table, path, with pandas'''
 
     @staticmethod
-    def table_name():
+    def table_name() -> str:
         '''Table name'''
         return 'path'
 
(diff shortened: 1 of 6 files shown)


claudious96 (approved):

suzieq/engines/pandas/path.py
```diff
@@ -137,7 +138,34 @@ def _init_dfs(self, ns, source, dest):
...
+            # See if we can find an mlag pair of devices that contains the SVI
+            # pandas has a bug that prevents us from using startswith with a
+            # tuple such as ("Vlan", "vlan", "irb.") directly instead of using
+            # the @svi_names trick
+            # pylint: disable=unused-variable
+            svi_names = tuple(["Vlan", "vlan", "irb."])  # noqa: F841
+            if ':' in source:
+                self._src_df = (
+                    self._if_df.query(
+                        f'@self._in_subnet-series("{source}", ip6AddressList)'
```
claudious96, line 150:
```suggestion
                        f'@self._in_subnet_series("{source}", ip6AddressList)'
```

ddutt: Nice catch. Fixed

Issue #925, opened 2024-03-01
[Bug]: Python API with namespace parameter


vsurresh:
### Suzieq version

0.22.0

### Install Type

hand deployed python

### Python version

3.9

### Impacted component

other

### Steps to Reproduce

If you use the python API to get path information using this code, for example, it gives an error.

```from suzieq.sqobjects import get_sqobject

path_tbl = get_sqobject('path')
path_df = path_tbl().get(src='1.1.1.1', dest='6.6.6.6', namespace='my_ns')

print(path_df)
```

The code expects the namespace argument as a **list** instead of a string. It would be good to get an error message that indicates this behaviour. This is the error we get at the moment.

```Traceback (most recent call last):
  File "/Users/sureshv/Documents/python/suzieq_project/venv/lib/python3.9/site-packages/suzieq/engines/pandas/path.py", line 54, in _init_dfs
    raise EmptyDataframeError
suzieq.shared.exceptions.EmptyDataframeError

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Users/sureshv/Documents/python/suzieq_project/path_example.py", line 4, in <module>
    path_df = path_tbl().get(src='1.1.1.1', dest='6.6.6.6', namespace='my_ns')
  File "/Users/sureshv/Documents/python/suzieq_project/venv/lib/python3.9/site-packages/suzieq/sqobjects/path.py", line 19, in get
    return super().get(**kwargs)
  File "/Users/sureshv/Documents/python/suzieq_project/venv/lib/python3.9/site-packages/suzieq/sqobjects/basicobj.py", line 226, in get
    result = self.engine.get(**kwargs, columns=columns)
  File "/Users/sureshv/Documents/python/suzieq_project/venv/lib/python3.9/site-packages/suzieq/engines/pandas/path.py", line 800, in get
    self._init_dfs(self.namespace, src, dest)
  File "/Users/sureshv/Documents/python/suzieq_project/venv/lib/python3.9/site-packages/suzieq/engines/pandas/path.py", line 56, in _init_dfs
    raise EmptyDataframeError(
suzieq.shared.exceptions.EmptyDataframeError: No interface information found for ['m']
```

### Expected Behavior

The script should work even if we pass it as a string or an error message that indicates this problem.

### Observed Behavior

we need to pass the namespace as a list

PR #932, opened 2024-04-04
Bug #925 fix

Sepehr-A:
## Bug #925  Report Summary

A user reported a usability issue with Suzieq version 0.22.0 when using the Python API to retrieve path information. The API call was made with `namespace` provided as a string instead of a list, leading to an `EmptyDataframeError`. The error did not clearly indicate the cause, leading to confusion.

## Changes Made

To address this issue, two significant changes were implemented:

### Introduction of Argument Validation in `basicobj.py`:

- A new method `_validate_list_args` was added to enforce that certain arguments, specifically `namespace` and `hostname`, are provided as lists. If these arguments are not lists, the method raises a `TypeError` with a clear and informative error message. 

### Modification in `topology.py`:

- The initialization of `self._namespaces` was set to an empty list `[]`. 
- In the `get` method, `self._namespaces` was changed to default to an empty list `[]` instead of using `self.ctxt.namespace`. This modification was implemented to address testing issues that arose following the introduction of `_validate_list_args` in `basicobj.py`.

### DIFF
--- a/suzieq/engines/pandas/topology.py
+++ b/suzieq/engines/pandas/topology.py
@@ -29,6 +29,7 @@ class TopologyObj(SqPandasEngine):
         self.lsdb = pd.DataFrame()
         self._a_df = pd.DataFrame()
         self._ip_table = pd.DataFrame()
+        self._namespaces = []
 
     @staticmethod
     def table_name():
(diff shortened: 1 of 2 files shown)


suzieq/engines/pandas/topology.py
```diff
@@ -77,7 +78,7 @@ class Services:
             extra_cols: list
             augment: any
 
-        self._namespaces = kwargs.get("namespace", self.ctxt.namespace)
+        self._namespaces = kwargs.get("namespace", [])
```
LucaNicosia, line 81: This is the wrong fix. You'll loose the content of `self.ctxt.namespace`. You should instead put `[]` only if both `kwargs` and ctxt namespace value is empty.
This will do the trick
``` suggestion 
        self._namespaces = kwargs.get("namespace", self.ctxt.namespace) or []
```

LucaNicosia (changes requested):
After you have done with the fixes. Squash all the commits on `basicobj` file

ddutt (approved):

Issue #929, opened 2024-03-25
[Bug]: AttributeError: 'DataFrame' object has no attribute 'ifState' when loading default dashboard


beufanet:
### Suzieq version

0.22.0

### Install Type

container

### Python version

whatever is bundled with your container

### Impacted component

suzieq-gui

### Steps to Reproduce

When you have no OSPF tables, when opening first dashboard of the GUI, I have errors : 

```
AttributeError: 'DataFrame' object has no attribute 'ifState'
Traceback:

File "/usr/local/lib/python3.8/site-packages/streamlit/scriptrunner/script_runner.py", line 557, in _run_script
    exec(code, module.__dict__)
File "/usr/local/lib/python3.8/site-packages/suzieq/gui/stlit/suzieq-gui.py", line 245, in <module>
    apprun()
File "/usr/local/lib/python3.8/site-packages/suzieq/gui/stlit/suzieq-gui.py", line 241, in apprun
    state.pages[page].build()
File "/usr/local/lib/python3.8/site-packages/suzieq/gui/stlit/status.py", line 33, in build
    self._render(layout)
File "/usr/local/lib/python3.8/site-packages/suzieq/gui/stlit/status.py", line 109, in _render
    self._chart_ospf(layout, ns_list)
File "/usr/local/lib/python3.8/site-packages/suzieq/gui/stlit/status.py", line 210, in _chart_ospf
    ospf_df['state'] = np.where(ospf_df.ifState == "adminDown",
File "/usr/local/lib/python3.8/site-packages/pandas/core/generic.py", line 5902, in __getattr__
    return object.__getattribute__(self, name)
```

### Expected Behavior

Perhaps we should have a way to disable OSPF tables either in the Poller or the GUI ? 

### Observed Behavior

This error is not shown when there is some passive OSPF interfaces configured (and so the OSPF table is not empty), that's why I'm thinking it comes from OSPF state.

ddutt:
Closed via PR #933

PR #933, opened 2024-04-05
Fixed issue ospf gui (Issue #929)

rickstardust:
A user reported an `AttributeError` occuring when using the main GUI page when having no OSPF tables.

## Related Issue

#929 

## Description

The error originated from the ospf engine: when merging the ospfNbr and ospfIf tables, the code didn't ensure that the schema was correct also in case of missing data. 

## Type of change

## Double Check

### DIFF
--- a/suzieq/engines/pandas/ospf.py
+++ b/suzieq/engines/pandas/ospf.py
@@ -103,11 +103,13 @@ class OspfObj(SqPandasEngine):
         nbr_df = self._get_table_sqobj('ospfNbr') \
                      .get(columns=nbrcols, **kwargs)
         if nbr_df.empty:
-            missing_cols = set(ospfschema.get_display_fields(['default'])) - set(df.columns)
+            ospf_clmns = ospfschema.get_display_fields(['default'])
+            missing_cols = set(ospf_clmns) - set(df.columns)
             for col in missing_cols:
                 df[col] = np.nan
             df = df.fillna({'peerIP': '-', 'peerHostname': '',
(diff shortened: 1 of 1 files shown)


suzieq/engines/pandas/ospf.py
```diff
@@ -102,6 +103,11 @@ def _get_combined_df(self, **kwargs):
         nbr_df = self._get_table_sqobj('ospfNbr') \
                      .get(columns=nbrcols, **kwargs)
         if nbr_df.empty:
+            missing_cols = set(ospfschema.get_display_fields(['default'])) - set(df.columns)
+            for col in missing_cols:
+                df[col] = np.nan
+            df = df.fillna({'peerIP': '-', 'peerHostname': '',
```
ddutt, line 109: Why fill peerIP with "-", and not "" like the others?

rickstardust: I wanted to play it safe and comply with preexisting code. Take a look at row 130 of the same file:
```
df = df.drop(columns=list(df.filter(regex='_y$|_x$')),
                     errors='ignore') \
            .fillna({'peerIP': '-', 'numChanges': 0,
                     'lastChangeTime': 0})
```
A similar behavior can be observed within the bgp.py and lldp.py files in the same directory. However, I think that, unless there's specific code around which wants "-" instead of "" to filter out data for those columns, "" should be used everywhere because it makes it easier to detect that the field is, in fact, empty. That may require changes which don't directly target this issue though.

PR #934, opened 2024-04-08
Feature nautobot inventory

### LINKED ISSUE #921: [Feature]: custom netbox queries / add nautobot 2.X support
### Suzieq version

0.22.0

### Install Type

container

### Feature type

Extend sq-poller

### Use case

Import devices from netbox based on arbitrary filters, not just tags. 
It also has the welcome side effect of adding support for Nautobot 2.X (the query syntax of netbox and nautobot 1.X is equivalent and so the netbox source can be used for nautobot deployments, but it has changed for nautobot 2.X)

### Proposed functionality/solution

The netbox source allow specifying one or more tags to filter which devices to include in the inventory.
The approach I'm proposing is to allow the SuzieQ user to specify the filters that will be sent to netbox.
Could be a list of tags, a list of sites, and/or a list of any other attributes...  

This makes the Netbox source more flexible, and also makes it work for all the Nautobot versions  

Example: this is a sample netbox source configuration
```
sources:
- name: nautobot
  token: env:NAUTOBOT_TOKEN
  url: env:NAUTOBOT_URL
  tag:
  - next_gen
  type: netbox
  period: 3600
```
and this is the equivalent config that uses the proposed filters mechanism
```
sources:
- name: nautobot
  token: env:NAUTOBOT_TOKEN
  url: env:NAUTOBOT_URL
  filters:
  - tag=next_gen
  type: netbox
  period: 3600
```
both produce an API request like this `GET https://demo.nautobot.com/api/dcim/devices/?tag=next_gen`

Another example that really shows the flexibility of this approach:
```
sources:
- name: nautobot
  token: env:NAUTOBOT_TOKEN
  url: env:NAUTOBOT_URL
  filters:
  - role=core
  - status=active
  - tags=next_gen
  - tags=whatever
  - depth=1
  type: netbox
  period: 3600
```
which produces an API call like `GET https://demo.nautobot.com/api/dcim/devices/?role=core&status=active&tags=next_gen&tags=whatever&depth=1`

It's up to the user to lookup the netbox/nautobot documentation and pick the parameters that make most sense for the use-case.

### External dependencies

_No response_

housepbass:

## Related Issue

Fixes #921 

## Description

Nautobot has been added as a dynamic inventory source for SuzieQ. 

## Type of change

- Documentation update

## New Behavior

Users can now specify `nautobot` as an inventory source for SuzieQ. The feature uses [Pynautobot](https://pynautobot.readthedocs.io/en/latest/) and allows for Nautobot 1.x and 2.x support. Filtering is possible using pynautobot's filters:

```yaml
sources:
  - name: nautobot-instance
    token: 0123456789abcdef0123456789abcdef01234567
    url: http://localhost:8080
    ssl-verify: false
    type: nautobot
    period: 3600
    device_filters: # pynautobot filters
      location: "Athens, GA"
      role: "Spine"
```

Nautobot is not currently supported.

- [] All my commits have `--signoff` applied (Only my last commit has this, how can I remediate?)


docs/inventory.md
```diff
@@ -3,8 +3,8 @@
 The inventory file that the poller uses describes a set of sources used to gather the list of devices, credentials to authenticate in the devices, default settings and eventually puts all together defining namespaces. An extensive explanation of each secation is provided on this page.
 
 !!! warning
-    Starting with version 0.16.0 the old options `-D` and `-a`  and the old inventory format are no longer supported.
-    See the section [Migrating to the new format](#migrating-to-new-format).
+Starting with version 0.16.0 the old options `-D` and `-a` and the old inventory format are no longer supported.
+See the section [Migrating to the new format](#migrating-to-new-format).
```
LucaNicosia, line 7: I think you have some kind of linter in your env. Can you disable it before updating the docs?

pyproject.toml
```diff
@@ -77,6 +79,7 @@ filelock = "^3.0.12"
 isort = "^5.10.1"
 pre-commit = "^2.19.0"
 toml = "^0.10.2"
+black = "^24.3.0"
```
LucaNicosia, line 82: Why did you added black as a dependency?

suzieq/poller/controller/source/nautobot.py
```diff
@@ -0,0 +1,228 @@
...
+
+class NautobotSourceModel(SourceModel):
+    """Nautobot source validation model"""
+
+    period: Optional[int] = Field(default=3600)
+    token: str
+    ssl_verify: Optional[bool] = Field(alias="ssl-verify")
+    server: Union[str, NautobotServerModel] = Field(alias="url")
+    run_once: Optional[bool] = Field(default=False, alias="run_once")
+    device_filters: Optional[Dict] = Field(default=None, alias="device_filters")
```
LucaNicosia, line 46: I don't see much validation on the `device_filters`. What happens if the user does the following?
``` yaml
- name: test
  type: nautobot
  device_filters:
    foo: bar
    bar: foo
```

suzieq/poller/controller/source/nautobot.py
```diff
@@ -0,0 +1,228 @@
...
+            List: inventory list
+        """
+
+        if not self._session:
+            self._init_session()
+        try:
+            if self._data.device_filters:
+                devices = self._session.dcim.devices.filter(**self._data.device_filters)
+            else:
+                devices = self._session.dcim.devices.all()
```
LucaNicosia, line 154: In netbox, this function handles NSOT servers behind a proxy. Do you know if the `pynautobot` automatically handles these kind of problems?

suzieq/poller/controller/source/nautobot.py
```diff
@@ -0,0 +1,228 @@
...
+            address = None
+
+            if ipv4:
+                address = ipv4
+            elif ipv6:
+                address = ipv6
+
+            if not address:
+                logger.warning(
+                    f"Skipping {namespace}.{hostname}: doesn't have a " "management IP"
```
LucaNicosia, line 194:
``` suggestion
                    f"Skipping {namespace}.{hostname}: doesn't have a management IP"
```

LucaNicosia (changes requested):

suzieq/poller/controller/source/nautobot.py
housepbass: Pynautobot will catch the error and an `InventorySourceError` will be raised with message `ERROR - nautobot-instance-123: error while getting devices: The request failed with code 400 Bad Request: {'foo': ['Unknown filter field']}`. I added a test for this as well.

docs/inventory.md
housepbass: Fixed.

suzieq/poller/controller/source/nautobot.py
housepbass: I'm not sure about that.

Issue #928, opened 2024-03-11
[Feature]: vSphere Support


netgun:
### Suzieq version

0.17.0

### Install Type

container

### Feature type

New component

### Use case

Why not support the Distributed Virtual Switch that exists within vSphere?  There could be a vast source of knowledge to be brought into SuzieQ.  And for those that run a Virtualized Datacenter without access to the physical devices, this could be huge

### Proposed functionality/solution

If you could read the Virtual Machine and associated network topology attached to it, you would have details on how all virtual machines are connected to the Virtual Switch level. 

### External dependencies

none

### Additional Context

_

ddutt:
Thank you for the suggestion @netgun . This is very useful, no doubt. But this will be a massive effort, as I see it. We have to acquaint ourselves with VMWare API, which we have no way to test, as we can't purchase VMW products :).

What help can you provide us?

tgupta3:
I'm happy to take this as we already have a lot of VMware deployments with & without NSX that I can use to test the code. In return, it would also be beneficial for us to observe the lifecycle of a VM. 

Are we just looking to discover what vms are connected to which dvs and which portgroup ?

ddutt:
What do you mean by "'m happy to take this on", @tgupta3 ?

tgupta3:
@ddutt  I meant I can implement this.

ddutt:
Fantastic @tgupta3 , happy to help you with this. @netgun if you can help test, that'd be great. 

First thing we need to do is define what is the information we want to save.

tgupta3:
IMO from troubleshooting point of view two things are very useful
1. If a VM existed at a point in time and if it was which segment and DVS it was on and also the vcenter it was on ?
2. The discovered IP for that VM, and maybe other metadata like creation date and template.

ddutt:
@tgupta3 So, what API calls are necessary for each? Whats the output of each of those API calls? We can start there

tgupta3:
I think [this](https://github.com/vmware/pyvmomi-community-samples/blob/master/samples/getvnicinfo.py) is a good example. It would look very similiar except use `content.viewManager.CreateContainerView` on `vim.DistributedVirtualSwitch`.

Once you have all the DVS, it's just a couple of for loops to iterate over the portgroups and then vms to get individual vm. 
[...]

PR #939, opened 2024-05-11
Add vcenter as inventory source

tgupta3:
## Related Issue

issue #928

## Description

Add vcenter support to suzieq. This will fetch the relevant VMs and their IPs that can be used to collect required data. The VM should have a custom attribute key (default is suzieq), the value of the key doesn't matter, in order for it to be part of the inventory.

Dependency added - pyvnomi

## Type of change

## New Behavior

A new vcenter source can be added like this
```
sources:
- name: vsphere-devices
  type: vcenter
  url: https://<url_vcenter>
  username: <trim>
  password: plain:<trim>
  ssl-verify: False
  attributes:
    - "monitored"
```

Vmware is commonly used in on-prem. This would allow monitoring of linux devices without specifying the devices individually. The change is backwards compatible. 

Add vcenter source for inventory. 

## Double Check

- [ x] I have read the comments and followed the [CONTRIBUTING.md](https://github.com/netenglabs/suzieq/blob/master/CONTRIBUTING.md).
- [ x] I have explained my PR according to the information in the comments or in a linked issue.
- [ x] My PR source branch is created from the `develop` branch.
- [ x] My PR targets the `develop` branch.
- [ x] All my commits have `--signoff` applied

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -52,8 +52,12 @@ packaging = "~21.3"
 psutil = "~5.9.4"
 jellyfish = "~0.10"
 altair = '>3.2, <5.0'
+<<<<<<< HEAD
 pydantic = '< 2.0'
 numpy = '~1.20'
+=======
+pyvmomi = "^8.0.2.0.1"
+>>>>>>> 5f7cedbc37 (Add vcenter as inventory source)
 
 [tool.poetry.dev-dependencies]
 pylint = "*"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -52,12 +52,9 @@ packaging = "~21.3"
 psutil = "~5.9.4"
 jellyfish = "~0.10"
 altair = '>3.2, <5.0'
-<<<<<<< HEAD
 pydantic = '< 2.0'
 numpy = '~1.20'
-=======
 pyvmomi = "^8.0.2.0.1"
->>>>>>> 5f7cedbc37 (Add vcenter as inventory source)
 
 [tool.poetry.dev-dependencies]
 pylint = "*"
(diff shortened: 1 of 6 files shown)


pyproject.toml
```diff
@@ -52,6 +52,7 @@ packaging = "^21.3"
 psutil = "^5.9.4"
 jellyfish = "~0.10"
 altair = '>3.2, <5.0'
+pyvmomi = "^8.0.2.0.1"
```
LucaNicosia, line 55: Hi @tgupta3. Thanks for the contribution! One question: we would like to keep the package as light as possible. Would it be possible for you to use the vcenter REST apis directly instead of the SDK? Look for example what we did in [netbox source](https://github.com/netenglabs/suzieq/blob/develop/suzieq/poller/controller/source/netbox.py)

tgupta3: Yeah, I have tried using the REST API ([Commit](https://github.com/tgupta3/suzieq/blob/vcenter/suzieq/poller/controller/source/vcenter.py)) but it's slow. The REST API has the following steps
1. Get a list of all VMs
2. For each vm in the list, do another API to get the ip address. 

It's not common to have 500 VMs in a single vcenter deployment, and a REST API call sometimes takes 1-2 seconds. With the SOAP SDK, it can be done in a single API call, and takes a few seconds vs few minutes with REST API.

ddutt (approved):

ddutt:
@tgupta3 Can you please rebase to latest deveop?

pyproject.toml
LucaNicosia: Thanks @tgupta3 . Sorry for the late response. I see your point and it makes sense. Leave the code as is

tgupta3:
> @tgupta3 Can you please rebase to latest deveop?

Done

claudious96 (approved):

PR #943, opened 2024-05-21
Eos routes wrong nexthopIps for protocol=connected entries

AndryNick98:
## Description
With eos devices, when we have an entry with protocol = connected we put 0.0.0.0 as nexthopIps while we should leave it empty.
The PR fix it

## Type of change

## Double Check

### DIFF
--- a/suzieq/poller/worker/services/routes.py
+++ b/suzieq/poller/worker/services/routes.py
@@ -48,6 +48,8 @@ class RoutesService(Service):
                 entry['oifs'] = len(entry['nexthopIps']) * \
                     ['_nexthopVrf:default']
             entry['protocol'] = entry['protocol'].lower()
+            if entry['protocol'] == 'connected':
+                entry['nexthopIps'] = []
             entry['preference'] = int(entry.get('preference', 0))
             entry['metric'] = int(entry.get('metric', 0))
             self._fix_ipvers(entry)


suzieq/poller/worker/services/routes.py
```diff
@@ -48,6 +48,8 @@ def _clean_eos_data(self, processed_data, _):
                 entry['oifs'] = len(entry['nexthopIps']) * \
                     ['_nexthopVrf:default']
             entry['protocol'] = entry['protocol'].lower()
+            if entry['protocol'] == 'connected':
```
LucaNicosia, line 51: In the commit you are saying that you should do this change only when 0.0.0.0 as nexthopIps. Here you are always updating it (hence you are not checking if `nexthopIps == '0.0.0.0'`. Is it the right thing to do?

AndryNick98: the point is that the nexthopIps list should be always empty if the protocol is connected. That was just an explanation of the bug, there are some cases where we put `0.0.0.0` as nexthopIps even if the protocol is connected, so here we just set nexthopIps as empty every time we have `protocol=connected` , it doesn't metter  checking if nexthopIps is equal to `0.0.0.0` or not.

ddutt (approved):

PR #1002, opened 2026-04-28
fix(cli): init rest config for per-command engine

### LINKED ISSUE #1000: [Bug]: per-command `engine=rest` does not initialize REST config from `suzieq-cfg.yml`
### Suzieq version

0.24.0

### Install Type

hand deployed python

### Python version

whatever is bundled with your container

### Impacted component

other

### Steps to Reproduce

Given a config like:

  ```yaml
  data-directory: ./my-data-dir

  rest:
    API_KEY: "my-api-key"
    no-https: True
    address: 127.0.0.1
    port: 8000
```
Start the CLI: `suzieq-cli -c ./suzieq-cfg.yml`

Then run: `sqPoller show engine=rest`

### Expected Behavior

When a command is invoked with `engine=rest`, SuzieQ should initialize the REST context from the loaded config
So this should work without requiring a prior global context change:  `sqPoller show engine=rest`

### Observed Behavior

The command fails with: `"ERROR: No REST API key specified"`

If the API key is manually set afterward:
```
set rest-api-key=my-api-key
sqPoller show engine=rest
```

the CLI still fails with the following error:
```
set rest-api-key=my-api-key
sqPoller show engine= rest
  Error: 1

  error
  0  ERROR: HTTPSConnectionPool(host='127.0.0.1', port=8000): Max retries exceeded with url: /api/v2/sqPoller/show?columns=default&view=latest&access_token=my-api-key (Caused by SSLError(SSLError(1, '[SSL:
  WRONG_VERSION_NUMBER] wrong version number (_ssl.c:1007)')))
```

### Additional Context

Running a CLI command with `engine=rest` fails unless the CLI was initialized with config including `ux.engine: rest`

#### Suspected Cause
The per-command engine path passes engine_name='rest' into the SqObject constructor, but it does not call set_rest_engine().

  Relevant code paths:
  - `suzieq/cli/sqcmds/command.py` passes the per-command engine into the SqObject.
  - `suzieq/sqobjects/basicobj.py` chooses engine_name directly when constructing the engine.
  - `suzieq/shared/context.py` only initializes REST settings when ux.engine: rest is loaded at context startup.
  - `suzieq/cli/sqcmds/context_commands.py` initializes REST settings only when running set engine=rest.

onthebed:
now pulls REST settings from the loaded config when the current context is still on pandas, so the CLI no longer needs a global engine switch first.

Added a unit test for the per-command rest path and verified the context gets the configured API key, host, port, and transport.

Fixes #1000
Pushed from BillionClaw.

### DIFF
--- a/suzieq/sqobjects/basicobj.py
+++ b/suzieq/sqobjects/basicobj.py
@@ -5,7 +5,8 @@ import pandas as pd
 from pandas.core.dtypes.dtypes import DatetimeTZDtype
 
 from suzieq.shared.utils import (load_sq_config, humanize_timestamp,
-                                 deprecated_table_function_warning)
+                                 deprecated_table_function_warning,
+                                 set_rest_engine)
 from suzieq.shared.schema import Schema, SchemaForTable
 from suzieq.engines import get_sqengine
 from suzieq.shared.sq_plugin import SqPlugin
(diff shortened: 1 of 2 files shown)


LucaNicosia (commented):
Hi @onthebed . Thanks for your contribution!

I tested this patch everything works!

Can you please signoff the commit? Once that's done, we can merge the PR

